Issue #5, opened 2019-10-25
develop tests for core functionality


davemfish:
At this point the project is stable enough that some of the core functionality should have tests before much more development happens.

davemfish:
Here's a list of user-interactions that should be simulated in tests, probably in a "snapshot" test framework:

- [ ] clicking on an invest model button to load it's setup form
- [ ] clicking on a recent session button to load a saved state
- [ ] entering invalid text in an input form to yield a validation message
- [ ] entering valid text in an input form to yield a valid signal
- [ ] using browse button to populate input field to an invalid state
- [ ] using browse button to populate input field to a valid state
- [ ] toggling radio buttons correctly set a boolean arg value
- [ ] batch load arguments from a logfile/parameter file to an invalid state
- [ ] batch load arguments from a logfile/parameter file to a valid state w/ execute button enabled
- [ ] batch load arguments from a file, overwriting existing form values
- [ ] batch load arguments from a file that does not correspond to the same invest model
- [ ] batch load arguments from a file that does not correspond to the same invest version 
- [ ] clicking execute button to save a JSON and trigger child process call to invest
- [ ] clicking execute displays LogTab with a logfile rendered
- [ ] Model Complete alert displays when execute finishes without errors
- [ ] Danger alert displays with stderr content when execute finishes with errors
- [ ] Save Session Modal Save - saves app state in JSON and adds to Recent Sessions buttons
- [ ] Save Session Modal Cancel - does nothing
- [ ] Save Parameters Button saves a JSON file as specified by natcap.invest.datastack.py
- [ ] Save Parameters Button is only enabled when SetupTab is populated with parameters
- [ ] Save Python Script Button saves a python file with an args dict and an execute call
- [ ] Save Python Script Button is only enabled when SetupTab is populated with parameters
- [ ] Settings Modal Save updates the app-level state for loggingLevel and nWorkers settings
- [ ] Settings Modal Cancel does not change app-level state, even after inputs have been touched.
- [ ] App-level invest settings (loggingLevel & nWorkers) are applied to investExecute calls and when saving JSON or Python scripts.
- [ ] Flask server reliably starts and stops when the Electron window opens and closes

davemfish:
Closing this issue. Surely there will be more test development in the future, but this issue was mainly useful for committing against during this first phase of test dev.

Issue #9, opened 2020-01-07
Load model inputs from the most recent session for that model by default


davemfish:
This is functionality we have in the Qt UI. 
E.g: When I select the Carbon model - populate the inputs with the values that were last used.

davemfish:
The "Recent Session" list arguably solves the same problem this feature would solve. To automatically populate a model's setup form with the same values used previously, select that previous "session" from the Home tab. 

There could still be good reason to implement the feature described in this issue though. Hopefully a little user-testing will make it clear.

davemfish:
I'll close this because the reason stated in the previous comment. And because initial user-testing has not generated any demand for changing these things.

Issue #6, opened 2019-11-08
Ordering, hiding, grouping of invest input fields


davemfish:
The UI layer populates an invest model input form given the model's `ARGS_SPEC`, but the UI should also be able to order the inputs in a coherent way for the user. Presently `ARGS_SPEC` does not include ordering information. 

Regardless of whether `ARGS_SPEC` is the place to keep this information, the UI requires knowing:

- a sensible order for the input fields, which vary by model
- whether some keys present in `ARGS_SPEC` should be hidden completely in the UI. For example, `recmodel_client` takes args for `hostname` and `port`, but we don't expose those in the UI.
- whether some inputs should be hidden until they are required. For example, several models (Rec, NDR, Hydropwer, et al) have boolean args, that when true, require other inputs. It's nice to keep the UI simple by hiding those other inputs until they become required.

davemfish:
I drafted a design doc for this issue. https://docs.google.com/document/d/1VYOoHYXV7gIf2VQaiphC9ahr8zOtbeETQRESQtyvEmg/edit#

Issue #3, opened 2019-10-25
Log display doesn't differentiate traceback from fatal vs. non-fatal stderr messages


davemfish:
For example, sometimes GDAL emits errors to stderr. Right now these are treated the same as a fatal invest exception, which means they get displayed as a scary alert (as if the model crashed) until the model completes and the alert is replaced with “Model Complete”.

PR #16, opened 2020-04-03
Task - test coverage and lots of core InVEST UI functionality

davemfish:
The goals for this PR were to reproduce the existing functionality of the python InVEST UI, and to have good documentation and test coverage.

Mainly this is an oppurtunity to get up to speed with the latest development of this project, and discuss and any changes of course that we might want to make.  

**Some things I feel good about and are less likely to change dramatically:**
* ReactJS for building the UI components
* leveraging existing functionality of `natcap.invest` (`validation`, loading model `ARGS_SPEC`, `datastack` functionality). Right now this happens with a Flask interface and that works quite well. But I don't think we ever really considered socket communication. So something to consider if we find shortcomings in the Flask approach.
* a node subprocess call to the invest CLI for running an invest model
* Jest + react-testing-library & dom-testing-library for integration and unit tests

**Some things that are more likely to change:**
* Packaging & distribution. Right now the app only works in a development setting. Some work has been done with packaging (see readme), but we have not tackled the python side of this at all.
* Somewhat related to above - the tests currently do not touch the python side (flask server or invest CLI). This is mainly to maintain speed and reliability of the tests. But we shouldn't ignore test coverage for this.
* Using a json file on disk as a persistent data store. This allows quick re-load of past sessions in a similar way that the current UI populates with values from a previous session by default. You'll notice the `jobdb.json` that gets created and added to anytime you run a model. 
* There are lots of places throughout the code with `TODO` flags and comments, but they are all fairly specific. 
* Almost everything related to style and appearance
* Anything related to open issues, especially natcap/invest-workbench#1,  natcap/invest-workbench#6, natcap/invest#874, 

**Getting setup:**
The readme is a good entry point for getting things setup. 

As a reminder, electron applications have two mostly separate javascript processes. The main process and the render process. In this project, `npm start` runs `main.js` (the main electron process) in a node environment. This process creates a browser window and loads `index.html`, which includes `index.js` as a script. `index.js` is the starting point of the render process, and it renders the root react component, `app.jsx`.

Please don't bother reviewing anything underneath `src/components/ResultsTab/Visualization/`. These are the model-specific visualizations and they only exist for Carbon and HRA. The HRA component is quite large and was developed in a separate project.

This PR should resolve issues natcap/invest-workbench#5 natcap/invest-workbench#3 natcap/invest-workbench#9 natcap/invest#876


src/server.py
```diff
@@ -44,10 +42,13 @@
     "sdr": "sdr.sdr",
     "seasonal_water_yield": "seasonal_water_yield.seasonal_water_yield",
     "urban_flood_risk_mitigation": "urban_flood_risk_mitigation",
+    "urban_cooling_model": "urban_cooling_model",
```
phargogh, line 45: I could be wrong about this, but this reminds me a lot of the `_MODEL_UIS` (or some name like that) dict in the CLI file, and it could make sense to reduce the duplication here since this is on the python side of things.  Would that make sense for this case?

tests/app.test.js
```diff
@@ -0,0 +1,93 @@
+import path from 'path';
+import fs from 'fs';
+import React from 'react';
+import { fireEvent, render, wait } from '@testing-library/react'
+import '@testing-library/jest-dom'
```
phargogh, line 5: What's the significance of the leading `@`?

phargogh (approved):
This is really exciting @davemfish !  I don't think I have any specific requests for things to change, just added a few comments as I had questions reading through things.  Overall, I think the project organization makes sense, I really like how the application and test logic is expressed throughout, and I can see how we might expand it down the road as needed!  I'm sure we'll trend towards some sort of preference or standard for things like imports, docstrings (or whatever the js equivalent is), etc, and I'd defer to you on what would be best.

Anyways, I added by 'approval' here, and very happy to talk more!

.env-example
```diff
@@ -1,10 +1,27 @@
...
+# to know whether we're in dev or production mode. Otherwise,
+# only the main process is aware of that. See index.js for use
+# of this variable in the renderer process.
 
+DEVMODE="true"
+
+# It's best to define the invest executeable path here, having it on your 
+# system's path already might work, but might not.
+
+INVEST="C:\\Users\\dmf\\Miniconda3\\envs\\invest-py37x\\Scripts\\invest.exe"
```
dcdenu4, line 14: For this file and line particularly, will this somehow be set by wherever the user Downloads InVEST to? Just curious how this file might look for packaging and deployment.

.gitignore
```diff
@@ -1,2 +1,8 @@
+build/
```
dcdenu4, line 1: I know there is much debate about including `.gitignore` in a repository or not. I'm fine with it, just wanted to make sure it was intended.

readme.md
```diff
@@ -4,8 +4,26 @@
```
dcdenu4, line 4: Does it make sense to have a little blurb at the top of the readme just generally describing what this project is and it's purpose? I realize no one will probably be finding themselves here for awhile but, might be nice.

src/InvestJob.jsx
```diff
@@ -38,32 +39,48 @@ if (process.env.GDAL_DATA) {
...
+        directory: null, suffix: null
+      },                               // only set values when execute starts the subprocess
+      logfile: null,                   // path to the invest logfile associated with invest job
+      logStdErr: null,                 // stderr data from the invest subprocess
+      sessionProgress: 'home',         // 'home', 'setup', 'log' - used on loadState to decide which tab to activate
+      jobStatus: null,                 // 'running', 'error', 'success'
+      activeTab: 'home'                // controls which tab is currently visible
     };
     
     this.argsToJsonFile = this.argsToJsonFile.bind(this);
```
dcdenu4, line 80: Hey Dave, could you make a comment about what these bind calls are doing? Not necessarily a code comment, but just a comment here for my curiosity. Do they help in handling callbacks or something?

src/InvestJob.jsx
```diff
@@ -73,59 +90,68 @@ export class InvestJob extends React.Component {
     this.setSessionID = this.setSessionID.bind(this);
   }
 
-  saveState(event) {
-    // Save a snapshot of this component's state to a JSON file.
+  saveState() {
```
dcdenu4, line 93: Ha, oh I think I get it relating to the bind comment.

src/InvestJob.jsx
```diff
@@ -136,261 +162,272 @@ export class InvestJob extends React.Component {
...
-      const progress = (code === 0 ? 'results' : 'log')
-      const workspace = {
-        directory: this.state.args.workspace_dir.value,
-        suffix: this.state.args.results_suffix.value
-      }
+      // TODO: there are non-zero exit cases that should be handled
+      // differently from one-another, but right now they are all exit code 1.
+      // E.g. this callback is designed with a model crash in mind, but not a fail to 
+      // launch, in which case the saveState call will probably crash.
+      const status = (code === 0 ? 'success' : 'error')
```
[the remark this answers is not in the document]
dcdenu4: This is interesting.

src/index.js
```diff
@@ -10,13 +12,34 @@ var _reactDom = _interopRequireDefault(require("react-dom"));
...
+const JOBS_DATABASE = 'jobdb.json'
+
+let rightClickPosition = null
+const menu = new Menu();
+menu.append(new MenuItem({
+  label: 'Inspect Element',
+  click: () => { 
+    remote.getCurrentWindow().inspectElement(rightClickPosition.x, rightClickPosition.y)
+  }
+}))
```
dcdenu4, line 29: Should something here have a semicolon? Actually above in lines 20, 21 there also are no semicolons.

src/server.py
```diff
@@ -57,17 +58,18 @@ def shutdown_server():
 
 @app.route('/ready', methods=['GET'])
 def get_is_ready():
-    return json.dumps('Flask ready')
+    return 'Flask ready'
```
dcdenu4, line 61: Should these tiny functions have docstrings still?

package.json
```diff
@@ -1,87 +1,94 @@
 {
   "name": "invest-electron",
   "productName": "invest-electron",
-  "version": "1.0.0",
+  "version": "0.1.0-alpha",
```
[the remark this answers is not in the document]
davemfish: Yes! `npm version` is the tool for that. So a release pipeline/checklist could include commands like:

`npm version patch`
`npm version minor` etc,

These automatically update the package.json version string -- according to semver specs -- and commit a git tag with the same string.

src/InvestJob.jsx
```diff
@@ -38,32 +39,48 @@ if (process.env.GDAL_DATA) {
...
 const TEMP_DIR = 'tmp'  // for saving datastack json files prior to investExecute
 
+// to translate to the invest CLI's verbosity flag:
+const LOGLEVELMAP = {
+  'DEBUG':   '--debug',
+  'INFO':    '-vvv',
+  'WARNING': '-vv',
+  'ERROR':   '-v',
+}
+
```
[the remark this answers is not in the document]
davemfish: I agree, I tend to prefer the more descriptive flags for command-line utilities. But I don't think it matters from the perspective of this UI.

src/InvestJob.jsx
```diff
@@ -73,59 +90,68 @@ export class InvestJob extends React.Component {
...
+    let job = {};
+    job[jobName] = {
+      model: this.state.modelName,
+      workspace: this.state.workspace,
+      statefile: filepath,
+      status: this.state.jobStatus,
+      humanTime: new Date().toLocaleString(),
+      systemTime: new Date().getTime(),
+    }
+    this.props.updateRecentSessions(job, this.props.appdata);
```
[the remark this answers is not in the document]
davemfish: That's right. `updateRecentSessions` will overwrite an existing entry with an identical `sessionID`, and that's intended since invest will also overwrite that model run. I think this okay for now, but if someday invest caches some interesting data that is not over-written when a workspace + suffix are re-used, then we could revisit this.

I'm pretty sure we're going to revisit the structure of this job metadata and the fact that it's just stored in a json file anyway. This is one of the most experimental parts of the UI at this point.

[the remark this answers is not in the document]
davemfish: Also in your example above: "what would happen if we re-run the same model in the same directory with the same suffix, but with different inputs?"

If a user did overwrite a previous run, but then wanted to redo that original run, they would no longer be able to get there by loading a "recent session", but they would still have the invest logfile from the earlier run and could load that via "Load Parameters" button.

src/InvestJob.jsx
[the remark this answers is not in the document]
davemfish: I made issue #18 to track this. I'll need to do some more testing to think about a good way forward. One problem which I sometimes hit during dev is an error from the shell -- not even from the CLI -- if `spawn` was given a bad `cmd`, for example.

src/InvestJob.jsx
```diff
@@ -136,261 +162,272 @@ export class InvestJob extends React.Component {
...
       model_module: this.state.modelSpec.module,
       args: args_dict_string
     };
+
+    // TODO: is there a use-case for `limit_to`? 
+    // Right now we're never calling validate with a limit_to,
+    // but we have an awful lot of logic here to cover it.
     if (limit_to) {
       payload['limit_to'] = limit_to
     }
```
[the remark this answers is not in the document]
davemfish: Thus far this new UI presents no reason to keep `limit_to`. There is issue #17 - related to runtime of `validate` functions. But I doubt the solution there is `limit_to`.

src/main.js
```diff
@@ -81,17 +82,15 @@ function createPythonProcess() {
...
-      if (!error) {
-        console.log('python killed')
-      } else {
-        console.log('Error: ' + error.message)
-      }
-    }
-  );
+function shutdownPythonProcess() {
+  return(
+    fetch('http://localhost:5000/shutdown', {
```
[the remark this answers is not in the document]
davemfish: Seems like a good idea to me!

src/server_requests.js
```diff
@@ -0,0 +1,126 @@
...
+// .then chaining of callbacks. Consider refactoring
+// everything here with async/await.
+
+export function getFlaskIsReady(retries=0) {
+  /* Recursive function to find out if the Flask server is online.
+  * Sometimes the app will make a server request before it's ready,
+  * so awaiting this response is one way to avoid that. 
+  */
+  return(
+    fetch('http://localhost:5000/ready', {
```
dcdenu4, line 13: I think James mentioned this being configurable somehow?sa

src/server.py
davemfish: Yes, I think that would work!

tests/app.test.js
davemfish: Good question. My reading is, `jest-dom` is the package, and it's developed under the scope of `@testing-library` project. Mainly that "scoping" has implications for npm. There's no other way to do the import here.

https://stackoverflow.com/questions/36667258/what-is-the-meaning-of-the-at-prefix-on-npm-packages

tests/utils.test.js
```diff
@@ -0,0 +1,133 @@
+import fs from 'fs';
+import path from 'path';
+import { findMostRecentLogfile, loadRecentSessions,
+         updateRecentSessions, boolStringToBoolean,
+         argsValuesFromSpec } from '../src/utils';
```
dcdenu4, line 5: I don't think `argsValuesFromSpec` is used anywhere?

dcdenu4 (approved):
Hey @davemfish this is some great work! And there's a lot of it. I put in some pretty small comments and questions. I think it will be great to have this officially tracked now and I'm looking forward to diving into pieces and parts as things come up. 

A few things I noticed:
- When selecting the 'X' close button in the app window the shell/terminal will hang on `Flask server shutting down...`
- Is the package-lock.json file something to be under version control?
- Seemed like there were some spots where semicolons weren't used, is there a JS best practice for semicolons? It's possible these spots I noticed don't actually need them, I get confused.

.env-example
davemfish: Good question, and part of larger discussion related to #1 . 

Right now, the `.env` file actually only loads in dev mode (i.e. the app is launched via `npm start`), not production mode. But I think it's a valid solution for production as well -- and we have no other solution for this yet, So we could generate and package a `.env` like this with the correct parameters set, depending on OS we're building for.

.gitignore
davemfish: I don't feel strongly either. I probably tracked it for my own convenience of developing on different machines. But this might be a good time to untrack it.

readme.md
davemfish: Yes, never hurts! I'll put this on my list for updating in this PR.

src/InvestJob.jsx
davemfish: Oh good! I'll explain it for myself anyway, just to convince myself that I also understand it. The `saveState` method wants to access properties of it's parent component (`InvestJob`). And so the method needs to be bound to that component. Once bound, `saveState`'s `this` object is also `InvestJob`'s `this`, so we can call `this.setState()` from inside the method, for example.

If not bound, `this` inside of `saveState` will not refer to the parent component's `this` at all.

src/index.js
davemfish: And I'm sure they're missing in many other places. I haven't been consistent. These days semicolons are added by the interpreter as needed, but there's still much debate. If we decide on a style guide we can go from there.

src/server.py
davemfish: Yes, will do.

davemfish:
> 
> A few things I noticed:
> 
>     * When selecting the 'X' close button in the app window the shell/terminal will hang on `Flask server shutting down...`
> 
Yeah I notice that too. Flask is shut down, so I think it's the electron/node process that's still not terminating. I'll look into that.

>     * Is the package-lock.json file something to be under version control?
It is, yes, but it's certainly not something whose changes should be reviewed. And it shouldn't be edited manually. It's useful because it describes dependenices with exact versions that npm has already determined are compatible by solving the more general requirements defined in `package.json`. So when the lock file is present `npm install` will avoid solving the dependency tree if it can - like if there's no new deps in `package.json`. That said, if the env gets messed up, it's sometimes useful to delete the `package-lock.json` and do `npm install`, which then refers only to the `package.json` (and re-makes a new `package-lock.json` in the process).
> 
>     * Seemed like there were some spots where semicolons weren't used, is there a JS best practice for semicolons? It's possible these spots I noticed don't actually need them, I get confused.
I think it's up to us to decide whether we want to code them, or let the interpreter insert them as needed. I've been inconsistent thus far.

davemfish:
Thanks for reviewing this @phargogh and @dcdenu4 ! So far I responded to the conversations you both started, and I've got a running list of items that you suggested (generally the small things) that I'm going to update while this PR is open. And then I added issues for the larger things, like #18 and #1. And I might do #17 in this PR also because it's pretty high priority.

davemfish:
Hey @phargogh and @dcdenu4 , I pushed changes with the following updates:

* configure port in .env
* replace the dict in server.py with cli._MODEL_UIS
* untrack .gitignore
* update Readme with general project desc/purpose
* docstrings in server.py
* added LICENSE from invest
* some linting
* properly terminating the app
* writeParamsToFile using datastack.py api, again. And test it.
* addressed the flaky not-fetching model list bug with a workaround

I think that's all I plan to do within this PR unless I hear otherwise from you!

src/InvestJob.jsx
```diff
@@ -136,261 +162,272 @@ export class InvestJob extends React.Component {
...
-        } else {
-          console.log('Status: ' + response.statusCode);
-          console.log('Error: ' + error.message);
-        }
-      }
-    );
+    
+    // TODO: is there much to be gained by outsourcing this to datastack.py via flask?
+    // It would insert the invest version string and enable an option to pass
+    // relative_paths=True. This is what that looks like:
```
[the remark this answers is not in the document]
davemfish: I switched this back to the `datastack.py` method in the latest changes. I think the reason I departed from that originally was to make testing easier, but I've gotten better at testing since then!

src/InvestJob.jsx
```diff
@@ -136,261 +162,272 @@ export class InvestJob extends React.Component {
...
-        sessionProgress: 'log',
-        activeTab: 'log',
-        logStdErr: '',
-        logStdOut: ''
-      }
-    );
+    // Write a temporary datastack json for passing as a command-line arg
+    const temp_dir = fs.mkdtempSync(path.join(process.cwd(), TEMP_DIR, 'data-'))
+    const datastackPath = path.join(temp_dir, 'datastack.json')
+    const _ = await this.argsToJsonFile(datastackPath);
```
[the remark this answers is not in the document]
davemfish: Yes, good suggestion. I created #19 to cover this. I think there are a couple other cases that should be configureable too.

dcdenu4:
@davemfish That sounds great to me! Thanks for addressing those and I'll let you or James do the honors of merging :)

package.json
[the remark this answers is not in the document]
phargogh: Very cool! `npm version` looks really replete

src/InvestJob.jsx
```diff
@@ -38,32 +39,48 @@ if (process.env.GDAL_DATA) {
...
+  * This component's state includes all the data needed to represent one invest
+  * job.
+  */
+
   constructor(props) {
     super(props);
 
     this.state = {
-      sessionID: defaultSessionID(''),
+      sessionID: null,                 // modelName + workspace.directory + workspace.suffix
```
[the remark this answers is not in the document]
phargogh: Resolved in another conversation on this PR: https://github.com/natcap/invest-gui/pull/16#discussion_r413196264

tests/app.test.js
phargogh: Ah!  That is a really interesting feature, and a bit surprising that it's implemented with some different import syntax.

phargogh:
Sounds great @davemfish!

PR #20, opened 2020-05-19
Refactor of the invest argument form - for performance and customization

### LINKED ISSUE #6: Ordering, hiding, grouping of invest input fields

### LINKED ISSUE #17: Input form updates are too sluggish for extremely fast keystrokes
I notice this if I type impossibly fast in an argument input text field, or if I lay on the backspace button. The text that is left in the field doesn't keep pace with the actual keystrokes.

The bottleneck here is the call to `investValidate` on every keystroke. I tested removing that call and the forms become fully responsive. Probably we don't need to validate the args on every keystroke, so a solution might be something like:

Hold the input form values in the local state of the `ArgsForm` component (right now they are passed down via props from `InvestJob`) and send them to validate at some frequent time interval.

davemfish:
This PR primarily addresses issues natcap/invest-workbench#17 & natcap/invest-workbench#6. 

**Ordering, grouping, and custom model forms:**
Model-specific data (a "UI Spec") is added to specify ordering and grouping of inputs on form, and to enable dynamic hiding or enabling/disabling of inputs that depend on values of other inputs. So far I only committed 3 such specs (`carbon, ndr, recreation`) to demonstrate the available features so far. Presence of a UI spec is not strictly required for a functioning user-interface.

**Refactors for performance and better seperation of concerns**
* All state related to invest model parameters was moved down to the `SetupTab` component where previously it was in `InvestJob`. 
* A new higher-order-component `ArgInput` is added to extract all the logic concerned with building individual input components for different types of invest args.
* Care is taken to minimize the amount of code running in `render` methods of `SetupTab` and `ArgsForm`, as these fire on every keystroke. 
* `investValidate` no longer updates the state object that controls the values of the inputs. It updates a separate state object that controls values of validation responses.


scripts/generate_base_ui_arg_specs.py
```diff
@@ -0,0 +1,48 @@
+"""
```
davemfish, line 1: No real need to review this unless you're curious.

src/components/HomeTab/index.jsx
```diff
@@ -23,7 +23,7 @@ const STATUS_COLOR_MAP = {
   success: 'rgba(40, 167, 69, 0.7)'
 }
 
-export class HomeTab extends React.Component {
+export class HomeTab extends React.PureComponent {
```
davemfish, line 26: Recommended reading: https://reactjs.org/docs/optimizing-performance.html#avoid-reconciliation

src/components/SetupTab/index.jsx
```diff
@@ -2,41 +2,354 @@ import React from 'react';
...
+
+    this.savePythonScript = this.savePythonScript.bind(this);
+    this.wrapArgsToJsonFile = this.wrapArgsToJsonFile.bind(this);
+    this.wrapInvestExecute = this.wrapInvestExecute.bind(this);
+    this.investValidate = this.investValidate.bind(this);
+    this.updateArgValues = this.updateArgValues.bind(this);
+    this.batchUpdateArgs = this.batchUpdateArgs.bind(this);
+  }
+
+  componentDidMount() {
```
davemfish, line 94: Recommended reading: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

src/InvestJob.jsx
```diff
@@ -143,30 +129,32 @@ export class InvestJob extends React.Component {
...
     /** Set this component's state to the object parsed from a JSON file.
     *
     * @params {string} sessionFilename - path to a JSON file.
     */
 
-    // const filename = path.join(CACHE_DIR, sessionFilename);
     if (fs.existsSync(sessionFilename)) {
       const loadedState = JSON.parse(fs.readFileSync(sessionFilename, 'utf8'));
+      
+      // Right now we saveState is only called w/in investExecute and only
```
dcdenu4, line 141: Typo: "Right now saveState is only ..."

src/InvestJob.jsx
```diff
@@ -143,30 +129,32 @@ export class InvestJob extends React.Component {
...
+      // after invest has created a logfile, which means an invest logfile
+      // should always exist and can be used to get args values and initialize SetupTab.
+      const datastack = await fetchDatastackFromFile(
```
dcdenu4, line 144: If the user accidentally deletes the log file or moves it around what happens to this call?

src/InvestJob.jsx
```diff
@@ -143,30 +129,32 @@ export class InvestJob extends React.Component {
...
+        { datastack_path: loadedState.logfile })
+      loadedState['argsInitDict'] = datastack['args']
+      Object.assign(loadedState, {setupKey: changeSetupKey(this.state.setupKey)})
```
dcdenu4, line 147: Okay, so changing this artificial `key` will cause `SetupTab` component to create a new component instance rather than update an old one. This will update all the values... Something like that?

src/InvestJob.jsx
```diff
@@ -276,103 +264,21 @@ export class InvestJob extends React.Component {
...
+  async investGetSpec(modelName, argsInitDict={}) {
     /** Get an invest model's ARGS_SPEC when a model button is clicked.
-    * A side-effect is that much of this component's state is reset.
+    *  
+    * Also get the model's UI spec if it exists.
+    * Then reset much of this component's state in case a prior job's 
+    * state exists. This includes setting a new setupKey, which is passed
+    * as a key to the SetupTab component, triggering it to re-mount
+    * rather than just re-render, allowing one-time initilization of
+    * arg grouping and ordering.
```
dcdenu4, line 275: Ah interesting. I wonder if it'd be useful to have an actual ID of some sort for every time a model button is clicked... maybe not.

src/InvestJob.jsx
```diff
@@ -276,103 +264,21 @@ export class InvestJob extends React.Component {
...
     *
-    * @param {string} - 
+    * @param {string} - as in a model name appearing in `invest list`
```
dcdenu4, line 277: I'm just noticing, does the argument / parameter name not need to be here? Is it order based?

tests/data/carbon_setup.json
```diff
@@ -1,5 +1,6 @@
 {
-  "sessionID": "carbon_setup",
+  "setupHash": "abfd5c3da293693a546c",
```
dcdenu4, line 2: I might have missed it, but what's this `setupHash`?

dcdenu4 (commented):
Hey @davemfish ,

I gave this a look over and added some questions and comments. Thanks for sharing the links to the reading materials. I don't have a ton of useful feedback, it all seems thought out and a good iteration. Perhaps at some point it will make sense spending a majority of a week really diving in so that I can be more useful on PRs such as this. 

The grouping, ordering, and functionality is exciting!

src/components/HomeTab/index.jsx
davemfish: That's correct. So if `props` have the same values as previous, this component will not even call `render`. `PureComponent` should be safe anytime the rendered DOM will be exactly the same given the same `props`. With a plain `Component`, `render` is always called whenever `props` or `state` change (which happens everytime a parent component called it's `render`), and then the new DOM is compared to the previous DOM to see if updates are actually needed.

src/InvestJob.jsx
davemfish: Yes! Creating the new component basically means calling SetupTab's `constructor` and `componentDidMount()`. That initializes some state and does the sorting and ordering of args. Basically all the stuff that needs to happen when a new model/ARGS_SPEC is selected, but not everytime the SetupTab renders (i.e. as user's interact with the form).

davemfish: It does seem like a bit of hacky way of controlling code execution, but it's one of the recommended patterns: https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key

src/InvestJob.jsx
davemfish: I think the main requirement is just that the key changes. At first I was re-generating a random string, then I thought I would hash the ARGS_SPEC + any initial arg values, then I ended up just incrementing an integer.

src/InvestJob.jsx
davemfish: Thanks for pointing this out. Documentation and style needs to be addressed throughout, I think. So I've created https://github.com/natcap/invest-gui/issues/21

But yes, the names should be listed there.

tests/data/carbon_setup.json
davemfish: Oops, this was an early version. I changed to `setupKey` in the end. I'll update this test data, or maybe see about generating it within the test itself.

src/InvestJob.jsx
davemfish: Great question. I'm adding some tests to make sure this is handled well.

davemfish: I added a check here to see if `datastack` exists and raise an alert message if not. But this also brings up the larger issue of designing how application state and sessions are stored on disk. All of that functionality is still kind of a prototype. I've created issue #22 .

davemfish:
Thanks for the review @dcdenu4 . I pushed a couple small changes and maybe more importantly, replied to comments and created issues #21 and #22 .

src/components/SetupTab/index.jsx
```diff
@@ -2,41 +2,354 @@ import React from 'react';
...
+        const group = Math.floor(argSpec.order)
+        let subArg = {}
+        if (argTree[group]) {  // already have arg(s) in this group
+          subArg[argSpec.order] = argkey
+          argTree[group].push(subArg)
+        } else {
+          subArg[argSpec.order] = argkey
+          argTree[group] = [subArg]
+        }
+      });
```
phargogh, line 133: Ah, very interesting grouping technique!

Do we have any 3rd-order groupings to worry about?

src/components/SetupTab/index.jsx
```diff
@@ -2,41 +2,354 @@ import React from 'react';
...
+      model_module: this.props.pyModuleName,
+      args: argsDictFromObject(argsValues)
+    };
+
+    // TODO: is there a use-case for `limit_to`? 
+    // Right now we're never calling validate with a limit_to,
+    // but we have an awful lot of logic here to cover it.
+    if (limit_to) {
+      payload['limit_to'] = limit_to
+    }
```
phargogh, line 246: I think we can safely remove anything that has to do with `limit_to`!  It's still used by the qt interface, but part of the intent of overhauling validation was to remove this optional parameter and all of the complexity that has to be put into place to support it.

src/components/SetupTab/index.jsx
```diff
@@ -65,30 +377,31 @@ class ArgsForm extends React.Component {
...
+  handleChange(event) {
+    /** Pass input value up to InvestJob for storage & validation.
+    *
+    * For text fields, this is on the onKeyUp handler, specifically 
+    * so that we don't call investValidate over and over while 
+    * laying on the backspace key (or any key).
+    */
+    const value = event.target.value;
+    const argkey = event.target.name;
+    this.props.updateArgValues(argkey, value)
```
phargogh, line 396: Ah, I was wondering how we'd be able to propagate changes back to a higher level!  Makes sense.

phargogh (approved):
Looks good to me, @davemfish !  I left a few comments, but don't have any major suggestions.  The recommended reading was also a great idea (not to mention very informative).

dcdenu4 (approved):
Thanks for addressing my questions Dave!

src/components/SetupTab/index.jsx
davemfish: Sounds good. I made #24 for this. I thought about doing it right here, but it's low priority and I'd rather not trigger a re-review of this PR.

davemfish:
Thanks @dcdenu4 and @phargogh for the comments. I'll go ahead and merge!

Issue #21, opened 2020-05-20
Decide on - and apply - a javascript style guide


davemfish:
Some resources to look into:

https://jsdoc.app/

https://google.github.io/styleguide/jsguide.html

davemfish:
We settled on the styles specified in `.eslintrc`

PR #27, opened 2020-06-30
Build and Distribution of the Electron package

davemfish:
This PR includes a github actions workflow using `electron-builder` to package our app into an executeable for Windows, MacOS, and Ubuntu. To support that workflow, other key changes and additions were made:

- New build scripts to collect the resources needed prior to packaging, including a configureable build of invest python binaries.
- Changes to how the local filesystem is used - mainly to make things work in production with the packaged app and in development with the src code, and to use the standard Electron "userData" directory. 
- Dependency on an invest binary build that includes this app's `server.py` module and it's dependencies (https://github.com/natcap/invest/pull/194)

Download artifacts: https://github.com/natcap/invest-gui/actions/runs/152210978

Bonus content:

- Some CSS work
- More test coverage, particularly around the child process calls to the invest CLI.


davemfish:
It looks like there are a couple issues with the Mac builds. 
1. There's both a dmg and zip that both contain the `.app`.  
2. There are spaces in the app's name that which screw up the path when trying to spawn the server and invest executeables.

Should be easy enough fixes. I'll push some changes here.

phargogh:
On the subject of dmg vs. zip, I'd opt for zipfiles as you wisely selected, if only because mac OS X will automatically unzip it for you and places it somewhere on the filesystem.  I've had _plenty_ of issues where folks would mount disk images and then not be able to run InVEST from the disk image.

package.json
```diff
@@ -6,21 +6,40 @@
   "main": "src/main.js",
   "homepage": "./",
   "scripts": {
-    "start": "cross-env FLASK_APP=src/server.py npx electron .",
+    "start": "cross-env FLASK_APP=src/server.py npx electron . --dev",
```
dcdenu4, line 9: Just curious, what's the `--dev` tag here?

src/index.js
```diff
@@ -1,7 +1,15 @@
...
 
-if (process.env.DEVMODE) {
+const isDevMode = function() {
+  return remote.process.argv[2] == '--dev'
+};
+if (isDevMode()) {
   require("@babel/register");
+  const dotenv = require('dotenv');
+  dotenv.config();  
 }
```
dcdenu4, line 13: Ah, found `--dev` mode. Could you just say what's happening here?

src/main.js
```diff
@@ -1,32 +1,61 @@
...
+
+// C) point to binaries included in this app's installation.
+} else {
+  const binary = (process.platform === 'win32') ? 'server.exe' : 'server'
+  // serverExe = path.join(__dirname, 'invest', binary)
+  console.log(process.resourcesPath)
+  serverExe = path.join(
+    process.resourcesPath, 'app.asar.unpacked', 'build', 'invest', binary)
+  console.log(serverExe)
+}
```
phargogh, line 39: This bit of InVEST binary loading looks pretty similar to what's in `src/index.js` and seems like an atomic operation, too ... would it be worth refactoring into a common place?

src/index.js
davemfish: Thanks, I added a comment. This loads the `.env` file that has some env vars pointing to your local invest binaries.

davemfish:
> Hey @davemfish , this looks like a lot of great work! I had a few curiosity questions but otherwise don't have a ton to contribute here at the moment. It looks like everything is moving in a good direction. I did see a few comments inside function blocks, I know you mentioned that those should go above? I didn't mark them because I wasn't sure if you wanted to deal with that here or had it's own issue open for it.

Thanks, Doug. We've got issue #21 for all that doc-formatting business.

src/main.js
davemfish: Very true. Instead of factoring out to a function in `utils.js` and calling it in both places, I changed this so `main.js` finds both the server and invest exe files (since they're always together) and then passes the necessary data to the renderer process (`index.js` and the rest of the app) via IPC. I bet that IPC could turn up useful for other stuff too.

davemfish:
Hey @phargogh I pushed some changes here. 
1. Removing the duplication that you mentioned. 
2. Addressing a `TODO` by moving some configuration out of a script and into `package.json` - where we define which invest version to build with and where on GCS to get the pre-built binaries.

That also reminds me, what do you think of the changes over in https://github.com/natcap/invest/pull/194 ? I don't think we should merge them too soon, but just don't want to forget that this PR depends on that stuff.

phargogh:
Ah, thanks for the reminder about https://github.com/natcap/invest/pull/194/files!  I think the changes proposed there look good to me and make sense for what we're trying to do

The PR is also making me wonder about a few things in the InVEST build process overall (not specific to this PR or build process, just noting them here for my own memory):

* The deleted hooks make me wonder if they're just no longer needed or if they actually conflict with the latest version of pyinstaller.  Should we upgrade pyinstaller in our build process?  Worth looking into.
* The binary build step for ubuntu looks pretty similar to the mac binary build step, and that makes me wonder if we'd be able to 1) move the windows build over to conda in order to 2) consolidate all our binary builds into a single configuration.
* The hardcoding of environment paths in our pyinstaller specs (which we've done forever) made me wonder if there were a way to pull the conda environement dynamically, maybe from `$CONDA_PREFIX` or something?
* Related to the hardcoding of environment paths, I wondered if the `gui` repo path in `pyinstaller.spec` could be pulled from the Makefile, something along the lines of `make jprint-GIT_GUI_REPO_PATH`

phargogh:
@davemfish what would you like to do about merging these PRs?  I think you mentioned that you wouldn't want to have https://github.com/natcap/invest/pull/194 merged until later on, but what about this one?

davemfish:
> @davemfish what would you like to do about merging these PRs? I think you mentioned that you wouldn't want to have [natcap/invest#194](https://github.com/natcap/invest/pull/194) merged until later on, but what about this one?

@phargogh this one could be merged whenever you approve. The invest one I would just close for the time-being and keep building invest from my fork as-needed.

davemfish:
> Ah, thanks for the reminder about [natcap/invest#194](https://github.com/natcap/invest/pull/194)! I think the changes proposed there look good to me and make sense for what we're trying to do
> 
> The PR is also making me wonder about a few things in the InVEST build process overall (not specific to this PR or build process, just noting them here for my own memory):
> 
>     * The deleted hooks make me wonder if they're just no longer needed or if they actually conflict with the latest version of pyinstaller.  Should we upgrade pyinstaller in our build process?  Worth looking into.
> 
>     * The binary build step for ubuntu looks pretty similar to the mac binary build step, and that makes me wonder if we'd be able to 1) move the windows build over to conda in order to 2) consolidate all our binary builds into a single configuration.
> 
>     * The hardcoding of environment paths in our pyinstaller specs (which we've done forever) made me wonder if there were a way to pull the conda environement dynamically, maybe from `$CONDA_PREFIX` or something?
> 
>     * Related to the hardcoding of environment paths, I wondered if the `gui` repo path in `pyinstaller.spec` could be pulled from the Makefile, something along the lines of `make jprint-GIT_GUI_REPO_PATH`

These are good ideas. I'm going to make a natcap/invest issue for the invest-gui support and I'll record some of these suggestions there.

PR #33, opened 2020-07-09
A logging module to log to console and files

### LINKED ISSUE #25: Log debug messages to a file in addition to the console
There are `console.log` calls throughout the source which will continue to be useful for debugging. But we should also log those messages to a file so that when the app is in production, we can have users submit bug reports with that log.

More broadly, manage logging so it is useful for bug reports.

davemfish:
I decided on `winston` for nodejs logging. https://github.com/winstonjs/winston. It was reasonably easy to setup and meets our needs. Though please see my comments within `logger.js` for some reservations. 

I also considered `pino`, but it's docs had such an emphasis on optimizing for the use-case of server-side logging handling millions of concurrent requests, and that's just not our use-case. That said, `winston` is also probably built with that case in mind.

The use-case I'm thinking of is mainly around troubleshooting and getting bug reports. Having logs in a file seems like a good first step for that. Logfiles go to Electron's "userData" directory.

closes #25


src/logger.js
```diff
@@ -0,0 +1,62 @@
...
+ * @returns {logger} - with File and Console transports. 
+ */
+function getLogger(label, filename='log.txt') {
+  // Right now the logging in this app has multiple loggers streaming to the
+  // same file. Maybe that's okay? The goal is having only one file on disk.
+  // But it feels wrong to create several different loggers just so that
+  // each message can have a 'label' that refers to the calling file.
+  // The only alternative I know of is to pass extra metadata anytime we log,
+  // but that also feels wrong:
+  // `logger.debug('foo', {label: filename})`
```
phargogh, line 31: Having multiple loggers all stream to the same file is a _very_common use case.  We do it all over the place in InVEST, for example!  Of course there is some extra metadata that we could include in the logfile if we ever really needed to differentiate between the sources of the log messages.  In any case, I'd expect `winston` to be able to handle concurrent access in a way that makes sense.

src/components/LogTab/index.jsx
```diff
@@ -81,7 +84,7 @@ export class LogTab extends React.Component {
         this.tail.unwatch()
       }
       catch(error) {
-        console.log(error)
+        logger.error(error.stack)
```
phargogh, line 87: Aaah nice.  I was wondering how winston would do something like `logger.exception()` in python.

src/main.js
```diff
@@ -69,7 +72,7 @@ const createWindow = async () => {
 
   // Create the browser window.
   const { width, height } = screen.getPrimaryDisplay().workAreaSize
-  console.log(width + ' ' + height)
+  logger.debug(width + ' ' + height)
```
phargogh, line 75: Is there a recommended way to do string replacement for log messages?  In other parts of this PR, for example, the `${variable}` notation is used instead.  Is one preferable for this use case?

I know we're only writing this to a logfile at the moment, but it could also behoove us to think forward to the possibility of sending log messages somewhere else.  In python's logging module, for example, the C-style '%s' notation was recommended to guard against code injection in the same way that MySQL queries support it as well.

src/main.js
```diff
@@ -111,24 +114,25 @@ function createPythonFlaskProcess(serverExe) {
...
     pythonServerProcess.stderr.on('data', (data) => {
-      console.log(`${data}`);
+      logger.debug(`${data}`);
     });
     pythonServerProcess.on('error', (err) => {
-      console.log('Process failed.');
-      console.log(err);
+      // If the python server process crashes, for now crash node also
+      logger.error(err.stack)
+      throw err
```
phargogh, line 128: Since the intent here is to crash node as well, could we add a note to the error message here about why this exception was triggered to be displayed in addition to the traceback?

And on a related note, when the python process does fail, will that process's exception be logged up on line 120 or 123 via stdout/stderr listening?

src/main.js
davemfish: Well this particular message can be nixed. But to answer your question, winston has C-style interpolation that can be enabled with a format option https://github.com/winstonjs/winston#string-interpolation

But elsewhere in the docs they say template strings are the way to go. https://github.com/winstonjs/winston#formats

What do you think?

src/main.js
davemfish: Can confirm that the python traceback is emitted and logged via stderr listening. In fact this `.on('error' )` listener may not come into play very often.

src/logger.js
davemfish: Great! In that case I pushed a change to use each file's filename as the logger's label to make reading the log a little easier.

src/main.js
davemfish: It looks like the 'error' listener is only used when the spawned process fails to start, or cannot be communicated with at all. I added a log message here as you suggest. 

Any excpetions from the subprocess go to the stderr listener. And in this case, with a flask process, flask is catching any uncaught python exceptions and re-raising them. So they do come through stderr, but the flask app also does not crash.

davemfish:
@phargogh, thanks for the review. I made various small changes here based on your feedback. 

And I addressed the unrelated problem of how to run `server.py` in dev mode without building binaries. After attempting some pyinstaller and setuptools options, I opted to just include a bit of extra logic in `main.js` and some extra configuration in `.env` so that `src/server.py` can be launched directly in dev mode. I did all that in this PR so I could experiment with raising exceptions from `server.py`

src/main.js
phargogh: For the sake of readability, I think the template strings are the way to go!

I think my concern with guarding against code injection is also a bit misguided as well and unnecessary for this use case since there's no opportunity for remote execution.  If someone injects code into the application, the worst that'll happen at this point is that they'll compromise their own computer.

src/logger.js
```diff
@@ -35,9 +36,12 @@ function getLogger(label, filename='log.txt') {
...
     })
 
-    const transport = new winston.transports.File({
+    const transport = new winston.transports.DailyRotateFile({
       level: 'debug',
-      filename: path.join(userDataPath, filename),
+      filename: path.join(userDataPath, 'invest-workbench-log-%DATE%.txt'),
+      datePattern: 'YYYY-MM-DD',
+      maxSize: '20m',
+      maxFiles: '3d',  // days
```
phargogh, line 38: Oh awesome ... I was wondering about this very thing!

phargogh:
Sounds great @davemfish!

PR #41, opened 2020-08-10
eslint, jsdocs, React PropTypes, and generally keeping things clean

davemfish:
This PR configures eslint (see `.eslintrc`) with extensions/plugins for the [airbnb style guide](https://airbnb.io/javascript/), React, and [JSDocs](https://jsdoc.app/). The airbnb style guide is very widely used. An alternative would have been [StandardJS](https://standardjs.com/). 

This PR touches nearly all the source code in order to standardize the syntax and language features we're using in this project. And some react components were refactored in small ways to avoid repitition or because their state was needlessly complex.

Tests were deliberately left out of the linting (for now) in order to avoid touching them more than necessary while other things were refactored. 

To improve test coverage at the start of this effort, I added `flaskapp.test.js` to cover `server_requests.js` and the flask endpoints themselves. Those tests deliberately do not mock the flask app, and so rely on the executeable that launches the flask app. That exe is not available in the CI that runs the other tests, so there is a new `npm run test-binaries` script to run these tests in the `build-electron.yml` instead of `run-tests.yml`. That's working well on Windows, but not on Mac - though the final Mac build is working.

And as part of sorting that out, `build-electron.yml` was re-configured to clone `natcap.invest` and build it's binaries using Pyinstaller and an `invest-flask.spec` script in this repo. So the workbench build process is no longer dependent on artifacts of the invest CI.


.github/workflows/build-electron.yml
```diff
@@ -7,29 +7,111 @@ jobs:
...
+      shell: bash -l {0}
+      env:
+        PIP_EXTRA_INDEX_URL: "http://pypi.naturalcapitalproject.org/simple/"
+        PIP_TRUSTED_HOST: "pypi.naturalcapitalproject.org"
+        PIP_PREFER_BINARY: 1
+        SITE_PACKAGES: "C:/hostedtoolcache/windows/Python/3.7.8/x64/lib/site-packages"
+      run: |
+        python -m pip install -r requirements.txt -r requirements-dev.txt
+        pip install flask
+        cp ${{ env.SITE_PACKAGES }}/shapely/DLLs/geos_c.dll ${{ env.SITE_PACKAGES }}/shapely/DLLs/geos.dll
```
dcdenu4, line 90: Is this shapely DLL dance related to Shapely 1.7.0+ ?

scripts/get_invest_binaries_gcs.js
```diff
@@ -1,3 +1,13 @@
+/**
+ * 2020-07-30 - the workbench build process no longer depends on
+ * pre-built binaries, instead building them itself, so this script
+ * is not being used anymore.
+ * 
+ * This is a utility script for fetching prebuilt invest binaries
+ * that include the server.py module and thus are compatible with the
```
dcdenu4, line 7: Is it worth keeping this file around, or can we just pull it back out from version control if we need it in the future?

.github/workflows/build-electron.yml
davemfish: @dcdenu4 , I think it goes way back. See also invest's `ci/windows-ci-binary-install.ps1` script.

src/InvestJob.jsx
```diff
@@ -12,131 +12,126 @@ import Navbar from 'react-bootstrap/Navbar';
...
+import {
+  getSpec, fetchDatastackFromFile, writeParametersToFile
+} from './server_requests';
+import {
+  findMostRecentLogfile, cleanupDir
+} from './utils';
 import { fileRegistry } from './constants';
-import { getLogger } from './logger'
-const logger = getLogger(__filename.split('/').slice(-1)[0])
+import { getLogger } from './logger';
```
dcdenu4, line 28: What do the brackets do here? I've read something indicating it depends on how they're exported and whether they are  a default or named export?

src/InvestJob.jsx
```diff
@@ -145,75 +140,86 @@ export class InvestJob extends React.Component {
...
 
+  /** Write an invest args JSON file for passing to invest cli.
+   *
+   * Outsourcing this to natcap.invest.datastack via flask ensures
+   * a compliant json with an invest version string.
+   *
+   * @param {string} datastackPath - path to a JSON file.
+   * @param {object} argsValues - the invest "args dictionary"
+   *   as a javascript object
+   */
```
dcdenu4, line 155: This looks like nice formatting

src/components/HomeTab/index.jsx
```diff
@@ -22,13 +21,14 @@ import Row from 'react-bootstrap/Row';
...
   running: 'rgba(23, 162, 184, 0.7)',
   error: 'rgba(220, 53, 69, 0.7)',
-  success: '#148F68' // invest green
-}
-
-export class HomeTab extends React.PureComponent {
-  /** Renders a button for each invest model and for each cached invest job.
-  */
-
+  success: '#148F68', // invest green
```
dcdenu4, line 24: InVEST Green! Should be a paint color you can find in stores, haha

src/components/SetupTab/ArgInput/index.jsx
```diff
@@ -8,159 +8,223 @@ import Row from 'react-bootstrap/Row';
...
+      argSpec,
+      handleBoolChange,
+      handleChange,
+      isValid,
+      selectFile,
+      touched,
+      ui_option,
+      validationMessage,
+      value,
+    } = this.props;
```
dcdenu4, line 52: I've noticed this new kind of set up in a few files. Is there anything you can comment on that might be interesting to me from a future developer point of view?

src/logger.js
```diff
@@ -3,58 +3,60 @@ const winston = require('winston');
...
 
-
 let userDataPath;
 let isDevMode;
 if (remote) {
   // When this module is imported from render process, access via remote
-  userDataPath = remote.app.getPath('userData')
-  isDevMode = remote.process.argv[2] == '--dev'
+  userDataPath = remote.app.getPath('userData');
+  isDevMode = remote.process.argv[2] === '--dev';
```
dcdenu4, line 11: Triple equals! Have always liked getting to use it.

.github/workflows/build-electron.yml
dcdenu4: Woah, I don't think I've ever looking in that file before. Thanks!

scripts/get_invest_binaries_gcs.js
davemfish: We could untrack it. That probably is the right thing to do at this point.

src/InvestJob.jsx
davemfish: Yes, exactly right. Since `logger.js` does not have a default export, but does export an object named `getLogger`, imports need to be "destructured" from the export object, which is what the braces do. In this case I think we could change this to a default export.

src/components/HomeTab/index.jsx
davemfish: Should probably also be a color we find in the CSS store instead of this jsx module! But I think that's business for another PR.

src/components/SetupTab/ArgInput/index.jsx
davemfish: This is [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). When it's done with `this.props` in a `render()` method, it's mainly about readability and avoiding all the extra `this.props` prefixes that would otherwise be needed to refer to each of these objects.

But destructring comes up in lots of other useful ways too, with examples in the MDN docs.

src/components/SetupTab/index.jsx
```diff
@@ -1,89 +1,83 @@
...
+import ArgsForm from './ArgsForm';
+import SaveFileButton from '../SaveFileButton';
+import { fetchValidation, saveToPython } from '../../server_requests';
+import { argsDictFromObject } from '../../utils';
+
+/** Toggle properties that control the display of argument inputs.
+ *
+ * This function returns a copy of the SetupTab argsValues object
+ * after updating the 'ui_option' property of any arguments listed
+ * in the 'ui_control' array of `argsSpec[argkey]`.
```
phargogh, line 19: Does `jsdoc` assume any particular markup language within docstrings?  Markdown?

src/components/SetupTab/index.jsx
```diff
@@ -207,286 +210,145 @@ export class SetupTab extends React.Component {
...
-      const argSpec = Object.assign({}, this.props.argsSpec[argkey])
-      if (argSpec.ui_control) {
-        argsValues = toggleDependentInputs(this.props.argsSpec, argsValues, argkey)
+    const { argsSpec } = this.props;
+    let { argsValues, argsValidation } = initializeArgValues(argsSpec, argsDict)
+    Object.keys(argsSpec).forEach((argkey) => {
+      if (argkey === 'n_workers') { return; }
+      const argumentSpec = Object.assign({}, argsSpec[argkey]);
+      if (argumentSpec.ui_control) {
+        argsValues = toggleDependentInputs(argsSpec, argsValues, argkey);
```
phargogh, line 219: This brings back memories of IUI's recursive toggling of state!

src/components/SetupTab/ArgInput/index.jsx
```diff
@@ -8,159 +8,223 @@ import Row from 'react-bootstrap/Row';
...
-          <Form.Label column sm="3"  htmlFor={argkey}>
-            {labelText}
-          </Form.Label>
+      const typeLabel = argSpec.type !== 'freestyle_string' ? argSpec.type : 'string';
+      Input = (
+        <Form.Group
+          as={Row}
+          key={argkey}
+          className={`arg-${ui_option}`}
+          data-testid={`group-${argkey}`}
```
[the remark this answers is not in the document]
davemfish: > Interesting! Is this parameter something that's only used in the tests? Or something else?

@phargogh That's right, it's only used in tests to select this DOM element and assert some content. It's meant as a last resort for selecting nodes, but there are a couple instances where the preferred selectors aren't viable. https://testing-library.com/docs/dom-testing-library/api-queries#bytestid

Here's an excerpt from `setuptab.test.js`

```
// The 'hide' style is applied to the whole Form.Group which 
// includes the Label and Input. Right now, the only good way
// to query the Form.Group node is using a data-testid property.
expect(utils.getByTestId('group-arg3')).toHaveClass('arg-hide')
```

src/components/SetupTab/index.jsx
davemfish: > Does `jsdoc` assume any particular markup language within docstrings? Markdown?

It looks like there's a plugin for Markdown rendered to html. But I haven't been concerned with any of that thus far. So far it doesn't feel like we're developing an API that would warrant rendered docs.

.github/workflows/build-electron.yml
```diff
@@ -40,6 +122,8 @@ jobs:
     
     - name: Upload installer artifacts
       uses: actions/upload-artifact@v2-preview
+      if: ${{ always() }}
```
emlys, line 125: Curious why the conditional is needed here?

src/components/SetupTab/ArgInput/index.jsx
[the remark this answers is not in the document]
emlys: Ooh! this brings back memories of UI testing with Cypress. It's weird to me that you have to add a property that's only used for tests, but I remember sometimes it is the only way. Tests that didn't access nodes by their test-id were often flaky, but hopefully that's less of a problem here. I like that you're assigning the `data-testid` programmatically rather than giving it an arbitrary name, that often was messy.

.github/workflows/build-electron.yml
davemfish: @emlys , I hope this is not needed long-term, but there's an issue with a previous step where tests using the binaries fail even though those same binaries work well in the final package. So I wanted to upload the build artifacts no matter what. 

In the long-term we'll definitely want to get to the root of those test failures, but it's also quite possible some part of this build/test process will get refactored for other reasons, so I didn't care to get it perfect for this PR.

PR #50, opened 2020-10-08
Terminating invest subprocesses | Automated browser tests

### LINKED ISSUE #36: Automated test of the binary builds
It would be useful to have some end-to-end tests of the distributions we're building with electron-builder. It is especially important because our jest test suite only tests the React side of the application and uses mocks for the electron API, the flask application, and the invest CLI subprocess. 

A framework like Spectron, or another inteface to chromedriver or webdriver, could run our packaged application in CI and give us confidence that the Node & Python layers of our application work together as expected.

https://www.electronjs.org/spectron
https://sites.google.com/a/chromium.org/chromedriver/
https://webdriver.io/

### LINKED ISSUE #46: Save Parameters button does not produce a viable datastack
The `args` property of the resulting json should only have the key:value pairs for the invest model's `execute` function. But right now `args` has all the "spec" data in it as well.

### LINKED ISSUE #47: Windows: invest runs popup in a new cmd shell
After setting up a run and clicking 'Execute', instead of switching to the workbench's Log tab display, Windows opens a new cmd window and invest runs there.

Probably we need to have a Windows-specific `child_process` call for launching invest to match the behavior we get with `spawn` on mac & linux.

davemfish:
This PR adds a feature to gracefully terminate an invest run during execution, with a cancel button. It also adds an automated browser test, using puppeteer to drive the browser, to test this functionality on the final built application.

I'm especially interested in review and advice on spawning and terminating subprocesses from node on different operating systems. This comes up in a few places:
* `src/InvestJob.jsx:investExecute` method. -- the invest model subprocess
* `src/main_helpers:createPythonFlaskProcess` function -- launching the flask app
* `tests/binary_tests/puppet.test.js` -- launching the built electron app as a spawned process

I believe the builds should all be successful, but `puppet.test.js` fails/errors in CI for various reasons:
* Windows: the test fails because on Windows we resort to force-killing the invest subprocess, which gives the app no time to handle the request to "Cancel Run" and display the proper message.
* MacOS: the test errors because: `Unable to create basic Accelerated OpenGL renderer. Core Image is now using the software OpenGL renderer. This will be slow.` I think this only comes up in the context of puppeteer, not when installing and launching the electron app. Would be great if @phargogh or @emlys could try running the app and this puppet test locally.
* Ubuntu: this was passing, but now seems unreliable, failing for various reasons. In CI, this test requires a virtual frame buffer provided by https://github.com/marketplace/actions/gabrielbb-xvfb-action. I'm not sure if that's a source of unreliability. The test is reliable outside of CI.

This PR also fixes #47, fixes #46, fixes #36


src/InvestJob.jsx
```diff
@@ -187,18 +184,42 @@ export default class InvestJob extends React.Component {
...
+      });
+      this.investRun.terminate = () => {
+        if (this.state.jobStatus === 'running') {
+          // process.kill(-this.investRun.pid, 'SIGTERM'); // does not kill
+          // this.investRun.kill(); // does not kill
+          // This kills, but no chance to handle it and show 'Run Cancelled'
+          exec(`taskkill /pid ${this.investRun.pid} /t /f`);
+          // exec(`taskkill /pid ${this.investRun.pid} /t`) // does not kill
+        }
+      };
```
dcdenu4, line 212: I did a little bit of looking into this and I probably found what you did; a lot of nodejs issues about how `process.kill` doesn't function correctly on Windows. It also seems like Windows doesn't use Signals in the same way.

The threads I was on: 
- https://github.com/nodejs/node/issues/3617
- https://github.com/nodejs/node/issues/12378

tests/app.test.js
```diff
@@ -460,4 +459,45 @@ describe('InVEST subprocess testing', () => {
...
+    // stdout listener is how the app knows the process started
+    mockInvestProc.stdout.push('hello from stdout');
+    const logTab = await findByText('Log');
+    expect(logTab.classList.contains('active')).toBeTruthy();
+    // some text from the logfile should be rendered:
+    expect(await findByText(dummyTextToLog, { exact: false }))
+      .toBeInTheDocument();
+    const cancelButton = await findByText('Cancel Run');
+    fireEvent.click(cancelButton);
+    expect(await findByText('Run Canceled')).toBeInTheDocument();
```
dcdenu4, line 493: So on Windows this will error because we can't properly `kill` a process?

tests/binary_tests/flaskapp.test.js
```diff
@@ -1,25 +1,26 @@
...
 beforeAll(async () => {
-	const binaries = await findInvestBinaries(true); // trues force devMode
-  createPythonFlaskProcess(binaries.server, true);
-  await server_requests.getFlaskIsReady()
+	const binaries = await findInvestBinaries(isDevMode);
+  createPythonFlaskProcess(binaries.server, isDevMode);
+  // In the CI the flask app takes more than 10x as long to startup.
+  // Especially so on macos.
+  // So, allowing many retries, especially because the error
+  // that is thrown if all retries fail is swallowed by jest
```
dcdenu4, line 17: graphic!

tests/app.test.js
davemfish: Yes. Tehcnically it fails because that "Run Canceled" message only shows when the subprocess has time to listen and run code on an `exit` event. Right now on Windows we're resorting to a force kill and there's no such time.

emlys:
Looks like it's not working on my mac because of the meddlesome space character in `~/Library/Application Support`! The full command ends up as: `invest -vvv run seasonal_water_yield --headless -d /Users/emily/Library/Application Support/invest-workbench/tmp/data-EYud8J/datastack.json`. Hence, an `invest: error: unrecognized arguments: Support/invest-workbench/tmp/data-EYud8J/datastack.json`.

davemfish:
> Looks like it's not working on my mac because of the meddlesome space character in `~/Library/Application Support`! The full command ends up as: `invest -vvv run seasonal_water_yield --headless -d /Users/emily/Library/Application Support/invest-workbench/tmp/data-EYud8J/datastack.json`. Hence, an `invest: error: unrecognized arguments: Support/invest-workbench/tmp/data-EYud8J/datastack.json`.

Aha! Great catch, @emlys ! Some quotes should do that job. I'll patch that up.

invest-flask.spec
```diff
@@ -48,17 +48,12 @@ invest_pyz = PYZ(invest_a.pure, invest_a.zipped_data, cipher=block_cipher)
...
-        (os.path.basename(name), name, 'BINARY') for name in
-        glob.glob(os.path.join(conda_env, 'lib/libspatialindex*.dylib'))]
-elif is_linux:
+if not is_win:
     # add rtree dependency dynamic libraries from conda environment
     invest_a.binaries += [
         (os.path.basename(name), name, 'BINARY') for name in
         glob.glob(os.path.join(conda_env, 'lib/libspatialindex*'))]
-elif is_win:
+else:
```
dcdenu4, line 56: Since we've made the switch to conda for the invest build workflows, could we add:

```python
# add rtree dependency dynamic libraries from conda environment
      invest_a.binaries += [
          (os.path.basename(name), name, 'BINARY') for name in
          glob.glob(os.path.join(conda_env, 'Library/bin/spatialindex*.dll'))]
```

This would also mean a change on line 9,10 to set `conda_env` for windows as well.

davemfish: (Hi, Rich)

phargogh (commented):
Looks good to me @davemfish !  That node behavior on process exit is pretty surprising and I do wonder about what alternatives/workarounds we might be able to implement, like we talked about on the coffee call this morning.  Maybe monitoring the process for cancellation would be best?  Or maybe we can just trust that cancellation will always go through?  I suppose if all else fails, there's always task manager.

tests/app.test.js
davemfish: @phargogh Maybe I mistated that. The callback is executed by the parent. Node (the parent) has a reference to the child process (`this.investRun`) that is listening for events from the child. So maybe the problem here is that the child never has the chance to emit the event?

src/InvestJob.jsx
```diff
@@ -185,20 +182,45 @@ export default class InvestJob extends React.Component {
...
+        env: { PATH: path.dirname(investExe) },
+        shell: true,
+      });
+      this.investRun.terminate = () => {
+        if (this.state.jobStatus === 'running') {
+          // process.kill(-this.investRun.pid, 'SIGTERM'); // does not kill
+          // this.investRun.kill(); // does not kill
+          // This kills, but no chance to handle it and show 'Run Cancelled'
+          exec(`taskkill /pid ${this.investRun.pid} /t /f`);
+          // exec(`taskkill /pid ${this.investRun.pid} /t`) // does not kill
```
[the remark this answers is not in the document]
davemfish: That's how I understand it as well.

emlys:
puppet.test.js still isn't working for me :( It's still failing on `expect(vals.includes('active')).toBeTruthy();`. `vals` is simply `nav-link disabled`. Everything works fine when I run it manually, and I can click on the Log tab right after clicking Execute. Maybe the assertion is happening too soon and it takes a moment for the Log tab to become active?

tests/app.test.js
davemfish: So now that I'm actually testing this on Windows, my last two statements are both wrong. The exit code on Windows after the force-kill is `1`.

davemfish:
> puppet.test.js still isn't working for me :( It's still failing on `expect(vals.includes('active')).toBeTruthy();`. `vals` is simply `nav-link disabled`. Everything works fine when I run it manually, and I can click on the Log tab right after clicking Execute. Maybe the assertion is happening too soon and it takes a moment for the Log tab to become active?

Thanks for continuing to test this @emlys . That's a good thought because the Log tab does not become active immediately, but I just double-checked and the test should be waiting and retrying that `expect` because it is wrapped in `waitFor` like this:

```
await waitFor(async () => {
    const prop = await logTab.getProperty('className');
    const vals = await prop.jsonValue();
    expect(vals.includes('active')).toBeTruthy();
  })
```

That same `Application Support/invest-workbench` folder should have logfiles that might reveal what's happening after the Execute button is clicked. Sometimes I `tail -f` that log and watch it while running the puppet test.

emlys:
Oh 🤦‍♀️ I needed to `npm run dist` again for the changes to take effect. The test is passing now!

davemfish:
Thanks for reviewing this, everyone. I pushed some changes to address the platform-dependent logic around killing the subprocess. I was partially mistaken when describing the problem earlier. The callback that listens for the `exit` event fires the same way for all OS. On Windows, the force kill yields an exit code of `1`, whereas on other systems the code was null. And that was the only real reason for the different behavior.

The only thing outstanding now is the Mac Build Action. I've made an issue for that (#52 ). Mac builds are good, it's just the puppeteer test that's not running properly in Actions.I don't want that to stop us from merging this PR though.

dcdenu4 (approved):

Issue #31, opened 2020-07-01
Define UI specifications for each invest model


davemfish:
Each model can have a JSON specification to define the relationships among individual input fields in that model's setup. The design doc for this feature: https://docs.google.com/document/d/1VYOoHYXV7gIf2VQaiphC9ahr8zOtbeETQRESQtyvEmg/edit#heading=h.9r56n017eqt4

Things that can be controlled:
* order of the arguments
* grouping of arguments
* argument dependencies - where the state of one argument controls whether it's dependent args are hidden/disabled.
* always-hidden arguments

A few models already have specs (e.g. https://github.com/natcap/invest-gui/blob/master/ui_data/natcap.invest.ndr.ndr.json). 

I can commit boilerplate specs for all other models, but then we'll need to populate those with values. These specs are not required for a functional application, but do provide a nicer user-experience.

davemfish:
- [ ] carbon                          
- [ ] coastal_blue_carbon             
- [ ] coastal_blue_carbon_preprocessor
- [ ] coastal_vulnerability           
- [ ] crop_production_percentile      
- [ ] crop_production_regression      
- [ ] delineateit                     
- [ ] finfish_aquaculture             
- [ ] fisheries                       
- [ ] fisheries_hst                   
- [ ] forest_carbon_edge_effect       
- [ ] globio                          
- [ ] habitat_quality                 
- [ ] habitat_risk_assessment         
- [ ] hydropower_water_yield          
- [ ] ndr                             
- [ ] pollination                     
- [ ] recreation                      
- [ ] routedem                        
- [ ] scenario_generator_proximity    
- [ ] scenic_quality                  
- [ ] sdr                             
- [ ] seasonal_water_yield            
- [ ] urban_cooling_model             
- [ ] urban_flood_risk_mitigation     
- [ ] wave_energy                     
- [ ] wind_energy

emlys:
Just started on this but before I get too far, what would you think about refactoring the UI spec `order` option a bit? Right now the order is determined by a decimal number where the integer part is the section, and the decimal part is the place within that section. E.g. `"order": 0.3` means the 3rd item in the 1st section (if I'm understanding correctly).

To me it might make more sense to have a separate `"order"` section in the top-level of the JSON. Something like: 
```
{
    "order": [
        ["workspace_dir", "results_suffix"],
        ["lulc_cur_path", "carbon_pools_path"],
[...]

PR #53, opened 2020-10-21
add uiSpec template files for all invest models that don't already have one

davemfish:
This is for #31 . Tracking files that are the `uiSpec` for each invest model. These will need editing to make them functional. This PR is just to start tracking the files.


davemfish:
We've already had a few models with a complete `uiSpec` before this PR. And all these files have been in my local repo directory for a long time. So I'm just going to merge this, knowing I'm not breaking anything or changing any functionality.

PR #59, opened 2020-10-27
Always use light mode

### LINKED ISSUE #58: Handle system light/dark mode settings
Observed on macOS, may apply to Windows as well?

System dark mode settings apparently apply to parts of apps by default. Opening InVEST with dark mode results in a black background with white text. It's alright but not aesthetically nice as the light mode version. The only real problem is that the log text ends up being invisible (black on black). It shouldn't be too hard to set a complementary dark mode color scheme.

emlys:
Fixes #58 
Override OS dark/light mode setting to always use light mode, until we get around to making a dark theme.
Also filled in the package.json description (copied from the invest repo)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "invest-workbench",
   "productName": "invest-workbench",
   "version": "0.1.0-alpha",
-  "description": "My Electron application description",
+  "description": "Models that map and value the goods and services from nature that sustain and fulfill human life",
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Hooray! First workbench PR!

I see this is branched from main so probably makes sense to merge it now and then I'll merge main back into the other larger pending PR.

PR #62, opened 2020-10-28
Fix SettingsModal console error

### LINKED ISSUE #55: React warning in the SettingsModal
Observed in `app.test.js` 

```
console.error
    Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

We should heed this warning.

As a bonus, I'd be delighted if we could figure out a good jest config to either A) cause these `console.error` to fail the test, or B) get jest to traceback all the way to the test file and line where this is raised. Here's the full traceback as of now. It's not super helpful in the context of `npm run test`.

```
  console.error
    Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
        in input (created by FormControl)
        in FormControl (created by SettingsModal)
        in div (created by Col)
        in Col (created by SettingsModal)
        in div (created by Row)
        in Row (created by FormGroup)
        in FormGroup (created by SettingsModal)
        in div (created by ModalBody)
        in ModalBody (created by SettingsModal)
        in form (created by Form)
        in Form (created by SettingsModal)
        in div (created by ModalDialog)
        in div (created by ModalDialog)
        in ModalDialog (created by Modal)
        in div (created by Modal)
        in Transition (created by Fade)
        in Fade (created by DialogTransition)
        in DialogTransition (created by Modal)
        in Modal (created by Modal)
        in Modal (created by SettingsModal)
        in SettingsModal (created by App)
        in nav (created by Navbar)
        in Navbar (created by App)
        in TabContainer (created by App)
        in App

      at printWarning (node_modules/react-dom/cjs/react-dom.development.js:88:30)
      at error (node_modules/react-dom/cjs/react-dom.development.js:60:5)
      at updateWrapper (node_modules/react-dom/cjs/react-dom.development.js:1870:7)
      at updateProperties (node_modules/react-dom/cjs/react-dom.development.js:6260:7)
      at commitUpdate (node_modules/react-dom/cjs/react-dom.development.js:7553:3)
      at commitWork (node_modules/react-dom/cjs/react-dom.development.js:20546:13)
      at commitMutationEffects (node_modules/react-dom/cjs/react-dom.development.js:22776:11)
```

emlys:
Fixes #55 
The `Warning: A component is changing an uncontrolled input of type text to be controlled` can happen when `setState` adds properties to the state. We should avoid it by making sure that all state properties that will ever be set are declared in the constructor.

The `<Form.Control>` that the error originates from is inside of a `<Modal>` and the error only appears in the test, not when using the app. However, taking it out of the `<Modal>` causes the error to show up when you start the app as well. This is kind of concerning... either the `<Modal>` is suppressing the error somehow, or the DOM being tested is different from that when you use the app?

I also added a jest config that causes tests to fail on `console.error` messages. I couldn't find any information about getting a more helpful traceback; other error tracebacks do have line numbers so maybe this is just an unusual situation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
   "jest": {
     "verbose": true,
     "collectCoverage": true,
-    "testEnvironment": "jest-environment-jsdom-sixteen"
+    "testEnvironment": "jest-environment-jsdom-sixteen",
+    "setupFiles": ["./tests/jest.overrides.js"]
   },
   "invest": {
     "version": "3.8.7"
--- a/src/components/SettingsModal/index.jsx
+++ b/src/components/SettingsModal/index.jsx
@@ -13,7 +13,10 @@ export default class SettingsModal extends React.Component {
     super(props);
     this.state = {
       show: false,
-      localSettings: {}
+      localSettings: {
+        nWorkers: '',
+        loggingLevel: ''
+      }
     };
 
     this.handleShow = this.handleShow.bind(this);
--- a/tests/jest.overrides.js
+++ b/tests/jest.overrides.js
@@ -1,0 +1,8 @@
+// Cause tests to fail on console.error messages
+// Taken from https://stackoverflow.com/questions/28615293/is-there-a-jest-config-that-will-fail-tests-on-console-warn/50584643#50584643
+let error = console.error
+
+console.error = function (message) {
+  error.apply(console, arguments) // keep default behaviour
+  throw (message instanceof Error ? message : new Error(message))
+}


davemfish:
> Fixes #55
> The `Warning: A component is changing an uncontrolled input of type text to be controlled` can happen when `setState` adds properties to the state. We should avoid it by making sure that all state properties that will ever be set are declared in the constructor.

Good reminder!

> The `<Form.Control>` that the error originates from is inside of a `<Modal>` and the error only appears in the test, not when using the app. However, taking it out of the `<Modal>` causes the error to show up when you start the app as well. This is kind of concerning... either the `<Modal>` is suppressing the error somehow, or the DOM being tested is different from that when you use the app?

Yes, slightly concerning. I'm pretty sure if the error was being raised we'd see it, and it is possible that the DOM is different between the test context and electron. Modals might be a special case in this respect. Tests run with jsdom. I think I had some related issues like this a while back, and then added the jest config `"testEnvironment": "jest-environment-jsdom-sixteen"`. I don't remember the specifics at this point. 

> 
> I also added a jest config that causes tests to fail on `console.error` messages. I couldn't find any information about getting a more helpful traceback; other error tracebacks do have line numbers so maybe this is just an unusual situation.

Much appreciated!

davemfish (approved):

PR #64, opened 2020-11-04
Separate concerns of an invest job persistent storage from the React UI

### LINKED ISSUE #57: separate the concerns of managing properties of invest jobs
I think we can improve the maintainability & readability of the codebase by making a class for invest jobs. This is different from the `InvestJob` React component, which manages the UI. This new class will manage the properties of an invest job and be responsible for setting & getting recent jobs from persistent storage. 

I've prototyped this already and it feels like a big improvement for maintainability & readibility.

davemfish:
This PR fixes #57 

It adds a new module to manage the metadata associated with an invest job, and to store and retrieve those metadata from a persistent database. For now we're using `localforage` to interface with the browser's built-in IndexedDB.

This PR also does some file renaming, which may be confusing for a second, but should be much less confusing in the long run.
* The new module is `src/InvestJob.js` and it exports the `InvestJob` class.
* The React component formerlly known as `src/InvestJob.jsx` is now `src/components/InvestTab/index.jsx` to match the convention of naming UI components after the thing they render (i.e. a Tab).
* Test files were renamed accordingly, so now we have `tests/investjob.test.js` and `tests/investtab.test.js`

This PR also stops ignoring the test files during linting. That was overdue and adds a lot of noise to the diff, but it's worth doing.


src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
+import crypto from 'crypto';
+import localforage from 'localforage';
+
+import { getLogger } from './logger';
+
+const logger = getLogger(__filename.split('/').slice(-1)[0]);
+
+const KEYS_ARRAY = 'sortedHashArray';
```
dcdenu4, line 8: Is `'sortedHashArray'` just the key name for the list of sorted keys?

src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /**
+   * @param {object} obj - the metadata property
+   * @param {string} obj.modelRunName - name to be passed to `invest run`
+   * @param {string} obj.modelHumanName - colloquial name of the invest model
+   * @param {object} obj.argsValues - an invest "args dict" with initial values
+   * @param {object} obj.workspace - defines the invest workspace and suffix
+   * @param {string} obj.workspace.directory - path to invest model workspace
+   * @param {string} obj.workspace.suffix - invest model results suffix
```
dcdenu4, line 46: Is it useful to have `suffix` nested under a `workspace` heirarchy? Any reason for not going with `obj.suffix` and `obj.workspace`?

src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
...
+   * @param {string} obj.logfile - path to an existing invest logfile
+   * @param {'running'|'success'|'error'} obj.status - status of the invest process
```
dcdenu4, line 48: Having this info is cool! Just out of curiosity what are the cases where I'd load a model back up and see `running`?

src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
...
+      );
+    }
+  }
+
+  setProperty(key, value) {
+    this.metadata[key] = value;
+    if (key === 'workspace') {
+      this.setWorkspaceHash();
+    }
+  }
```
dcdenu4, line 94: I think I'm a little confused with how the `workspace` is used. So if we have `key` of `'workspace'`, would the `value` coming in be an object with `suffix` and `directory`?

src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
...
+    // and make sure to move it to the front
+    const idx = sortedKeys.indexOf(this.metadata.workspaceHash);
+    if (idx > -1) {
+      sortedKeys.splice(idx, 1);
+    }
+    sortedKeys.unshift(this.metadata.workspaceHash);
+    localforage.setItem(KEYS_ARRAY, sortedKeys);
+    localforage.setItem(
+      this.metadata.workspaceHash, this.metadata
+    );
```
dcdenu4, line 116: Would it be a little nicer to do:
```javascript
if (! idx > -1) {
  sortedKeys.unshift(this.metatdata.workspaceHash);
}
localforage.setItem(KEYS_ARRAY, sortedKeys);
.....
```
Might be able to remove the `splice`?

dcdenu4 (commented):
Hey @davemfish this seems great. I had a few questions and comments, but they're probably all from lack of being familiar with the project and JS. 

Would it be helpful to test this locally on machines? If so, I'm happy to do that.

src/InvestJob.js
davemfish: Yeah. That's a pretty ugly name, huh. Both the variable and the string. Maybe it would be better like
`const HASH_ARRAY_KEY = 'workspaceHashes'

src/InvestJob.js
davemfish: Not a very important reason, no. But my reasoning was that it's really the directory & the suffix that together uniquely identify an invest "workspace". So that's why I had them bundled in one object.

At this point it may not even be useful to have them stored anywhere besides inside of `argsValues`. I think I should re-visit that before we merge because storing the same data in multiple places, without a good reason, seems like a bad idea.

src/InvestJob.js
davemfish: The status is updated when the run button is clicked (to 'running') and then again when invest exits (to either 'sucess' or 'error'). And the `save()` method should be called at both of those times as well. So if you were to switch back to the Home Tab while the model was running, the card should display with a spinner.

But, I just tested this and I think I have found a related bug that I introduced in this refactor. So I'd better fix that!

src/InvestJob.js
davemfish: That's right. But based on your earlier comment, I think I agree that might be a bad design. 

And the larger idea is - "an invest job with a workspace" is the "unit" worth saving to the recent jobs list. As opposed to an invest model setup without a workspace - since that model will not yet have been run.

src/InvestJob.js
davemfish: So if it **does not** already exist, add it to the front. But if it **does** already exist, we want to "move" it to the front in order to keep the array sorted (most recently saved first). And I think the way to do that is in two steps: delete the element with `splice` and then add it to the front with `unshift`.

davemfish:
> Hey @davemfish this seems great. I had a few questions and comments, but they're probably all from lack of being familiar with the project and JS.
> 
> Would it be helpful to test this locally on machines? If so, I'm happy to do that.

Thanks for your comments @dcdenu4 ! I think I found some bugs when I looked into them, so I'll have more changes coming. And yes, always useful to test at this stage! Though if you want to wait for the next changes that might be a good idea.

src/InvestJob.js
emlys: The `splice` then `unshift` makes sense to me. Though why they decided to name it `unshift` does not 🙄

src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
...
+const KEYS_ARRAY = 'sortedHashArray';
+
+/**
+ * Create an object to hold properties associated with an Invest Job.
+ *
+ */
+export default class InvestJob {
+  /* If none exists, init an empty array for the sorted workspace hashes */
+  static async initDB() {
+    const keys = await localforage.getItem(KEYS_ARRAY);
```
emlys, line 17: This is cool! To make sure I'm understanding...  because Electron uses Chromium (?) it's supported? And the benefit of `localforage` over `localStorage` is that you can access the stored data asynchronously?

src/InvestJob.js
```diff
@@ -0,0 +1,119 @@
...
+    if (!keys) {
+      localforage.setItem(KEYS_ARRAY, []);
+    }
+  }
+
+  /* Return an array of job metadata objects, ordered by most recently saved */
+  static async getJobStore() {
```
emlys, line 24: Is the size of the job store limited to a certain number of jobs, or is `clearStore` ever automatically called? It seems like this could build up pretty fast

src/InvestJob.js
davemfish: Tangentially related to this, I removed the Spinner - I find them slightly frustrating to have on the screen - and replaced with text of the status itself if it is 'running' or 'error'.

src/InvestJob.js
davemfish: Yep, we have full use of Chromium! And yes, `localforage` supports async and `localStorage` does not. That's pretty important, although our data is very small at this point, so synchronous calls might be okay. 

But the decision to go with `localForage` was more about this:

`localStorage` only supports simple key: value pair data, where the value is a string. We have more structured data we want to store, like actual javascript objects. It could work to wrap everything in `JSON.stringify` and `JSON.parse` in order to meet that localStorage constraint.

But, all modern browsers also support IndexedDB, which is a full-fledged database that can store our objects as-is. IndexedDB might be overkill for us though (at least at this point) and it has a much lower-level API, so `localForage` is a useful wrapper for both storage options. It stores our data with IndexedDB, but has a really simple API just like `localStorage`.

https://localforage.github.io/localForage/

It's such a simple API that it seemed like the right place to start. There will be very little to change if we decide to replace with a different database at any point.

src/InvestJob.js
davemfish: Good point! I added a max of 30 jobs for now - always dropping the oldest one when a new one is saved after that. And also added a "clear" button in the settings dialog.

davemfish:
Thanks for your reviews @emlys and @dcdenu4 . I pushed some new changes - you can see what's changed since your last review here: https://github.com/natcap/invest-workbench/compare/438fa21..d55d63c

* Removed some redundancy from the `metadata` property of an `InvestJob` instance - specifically the `workspace` property. 
* Limited the number of jobs that can be stored in the database, and added a "clear" button
* Fixed a bug that allowed the "Cancel Run" button to be enabled even if there was no running invest process
* Fixed a bug where the invest log would not refresh if a mdoel was being re-run

Back to you!

dcdenu4 (approved):
Thanks @davemfish ! Just did a pass over the changes you made. Looks pretty good to me!

src/InvestJob.js
emlys: Cool, thanks for the explanation. It does sound like just the right thing for our purpose!

emlys (approved):
Looks good!

Issue #61, opened 2020-10-28
Model Validation: distinguish between required & optional inputs that have not yet been touched


davemfish:
We don't display the validation feedback for inputs until after they have been "touched". That's to prevent many scary invalid messages from appearing right away when an empty form is loaded. But it also means the user can complete nearly all of the inputs, but still not have the run button enabled, but also not have any feedback about which fields still require an input (vs. which are okay to leave blank).

davemfish:
A good solution here - following the Qt UI - could be to always show the :heavy_check_mark: or :x: whether or not the field has been touched, but only display the detailed feedback message for invalid fields that have been touched. Or even only show the feedback while the field is in "focus".

To be clear on the original issue, validation is always happening, we're just being selective - maybe too selective - about when the results are being displayed to the user.

davemfish:
Fixed by #68

PR #68, opened 2020-11-18
Bugfix and cleanup for the validation feedback display

### LINKED ISSUE #66: Model Validation: handle validation feedback that applies to multiple inputs
The canonical example is the "Bounding boxes do not intersect" message. It's very verbose and, as-is, the message is duplicated below each of the inputs involved in the overlap check.

We should reduce the clutter of the duplicated messages by finding one single place to display the feedback message, while still indicating which inputs are involved.

davemfish:
This Fixes #66 and #61 

Invest validation results are displayed as follows:
* If form is loaded with initial inputs (from datastack/logfile or recent job) all validation feedback is displayed on all inputs immediately
* If a blank setup form is loaded for a model:
  + Validation runs and :x: is displayed, next to invalid inputs. With no values yet entered, required inputs will be invalid, optional inputs will be valid. This fixes #66 .
  + The feedback message for a field only appears once the field has been touched (either a 'change' or a 'focus'). Similarly, the :heavy_check_mark: does not appear on optional inputs until they are touched.

This is a compromise between not overloading the UI with validation feedback immediately when the blank form is loaded, but still differentiating between required & optional inputs so it's always clear which input(s) still need to go valid in order to enable the Run button.

For #61, the Bounding Box feedback message is formatted like this. Each input only reports on it's own bounding box. And the boxes are arranged with fixed-width to help with comparing the boxes and finding the one(s) that don't match the others.
Screenshot from 2020-11-18 05-39-37

### DIFF
--- a/src/components/SetupTab/ArgInput/index.jsx
+++ b/src/components/SetupTab/ArgInput/index.jsx
@@ -9,6 +9,20 @@ import Row from 'react-bootstrap/Row';
 import InputGroup from 'react-bootstrap/InputGroup';
 import Modal from 'react-bootstrap/Modal';
 
+/**
+ * Filter a message that refers to many spatial inputs' bounding boxes.
+ *
+ * Messages are oneline strings with `|` separating each
+ * filename & bounding-box pair. E.g:
+ *
+ *   `Bounding boxes do not intersect:
+ *   ./a.shp: [-84.9, 19.1, -69.15, 29.5] |
+ *   ./b.shp: [-79.0198, 26.481, -78.3717, 27.268] | ...`
+ *
+ * @param {string} message - a string that contains `filepath` e.g:
+ * @param {string} filepath - string preceding the relevant part of `message`
+ * @returns {string} - the filtered and formatted part of the message
+ */
 function filterSpatialOverlapFeedback(message, filepath) {
   const newPrefix = 'Bounding box does not intersect at least one other:';
   const bbox = message.split(`${filepath}:`).pop().split('|')[0];
--- a/src/components/SetupTab/ArgInput/index.jsx
+++ b/src/components/SetupTab/ArgInput/index.jsx
@@ -1,3 +1,4 @@
+import os from 'os';
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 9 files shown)


src/components/SetupTab/ArgInput/index.jsx
```diff
@@ -8,6 +9,29 @@ import Row from 'react-bootstrap/Row';
...
+ *
+ * @param {string} message - a string that contains `filepath` e.g:
+ * @param {string} filepath - string preceding the relevant part of `message`
+ * @returns {string} - the filtered and formatted part of the message
+ */
+function filterSpatialOverlapFeedback(message, filepath) {
+  const newPrefix = 'Bounding box does not intersect at least one other:';
+  const bbox = message.split(`${filepath}:`).pop().split('|')[0];
+  const bboxFormatted = bbox.split(' ').map(
+    (str) => str.padEnd(22, ' ')
```
phargogh, line 30: Nice, I was wondering about how the padding would be implemented!  How did you arrive at 22, and what will happen if we overrun the 22-character padding here?

src/components/SetupTab/ArgsForm/index.jsx
```diff
@@ -1,3 +1,4 @@
+import os from 'os';
```
phargogh, line 1: Bonus import?

src/styles/style.css
```diff
@@ -248,7 +248,11 @@ input[type=text]::placeholder {
 }
 
 .invalid-feedback {
-	font-size: 1.05rem;
+	font-size: 1.15rem;
+	font-family: monospace;
+	white-space: pre-wrap;
+	padding-left: 50px;
+	text-indent: -50px;
```
phargogh, line 255: I'm sorry I'm not super well-versed in the tradeoffs, but I noticed that a lot of the rest of the sizes in this file are defined in `rem` rather than `px` ... should these be in `rem` as well?

phargogh (commented):
Nice!  I particularly like the care taken to align the bounding box coordinates in the error messages.  Just had a couple of small questions (hence the 'comment') status, and I'm not sure whether they're a big deal and wanted to ask.  Let me know what you think, and happy to talk more and re-review!

src/components/SetupTab/ArgInput/index.jsx
davemfish: I had some "real" sample data coordinates with ~18 digits, so 22 allowed for those and added some extra space. If the input string is longer than the pad length, it's returned unchanged. 

I also thought about first reducing the precision to 7 or 8 places, then padding, for even better readability. But I didn't end up bothering with that. I still think it's a good idea since these coords are always returned in decimal degrees, so there's tons of false precision here. But maybe it's better to make that change in `validation.py`, if we want to.

src/components/SetupTab/ArgsForm/index.jsx
davemfish: Thank you!

src/styles/style.css
davemfish: Good point! This is ta rick to get a "hanging indent", which isn't critical at all, but I think its nice for these multiline messages.

davemfish:
Good points, James. Thanks for the review! I changed the `px` units to `rem` and removed the unused import. Back to you @phargogh

phargogh (approved):
Looks great ... thanks @davemfish !

Issue #52, opened 2020-10-16
MacOS puppeteer test failing


davemfish:
`puppet.test.js` errors in Actions on MacOS for reasons that appear to be Actions-specific. The error does not reproduce locally. It might be related to the fact that there is no display available in Actions.

Here's a failing test:
https://github.com/natcap/invest-workbench/pull/50/checks?check_run_id=1266102726

This message is concerning, and it's something that only comes up in Actions.
```
console.log
    Unable to create basic Accelerated OpenGL renderer.
    Unable to create basic Accelerated OpenGL renderer.
    Core Image is now using the software OpenGL renderer. This will be slow.

      at Socket.<anonymous> (tests/binary_tests/puppet.test.js:67:13)
```

PR #72, opened 2020-11-30
move python side of workbench to the invest repo

### LINKED ISSUE #70: move python side of workbench to natcap.invest repo
This issue corresponds with https://github.com/natcap/invest/issues/341

For the workbench, it means,

untracking `server.py`
untracking the pyinstaller script
downloading pre-built invest binaries during the workbench build process instead of building them from source
launching the flask app from the single invest executeable created from natcap.invest.cli

### LINKED ISSUE #71: Drop binary build support for linux
With the python side of this project moving to invest (#70 ) - binary build support for linux would need to move there as well. That would be new for invest, and while there's some demand (Martin & me) it's not something we need to support right away. 

The main obstacle right now is this issue with shapely, conda-forge, & PyInstaller. https://github.com/Toblerity/Shapely/issues/831

I'd rather wait for the upstream fixes than try to patch something in with custom pyinstaller hooks that will only be temporary.

Linux developers can still develop and run the workbench by binding to invest through an invest CLI in a local python env, rather than binding to frozen binaries from pyinstaller. See the `.env-example` file.

davemfish:
This PR fixes #70 and fixes #71 

The main changes are to the build process. The workbench build process will no longer build invest binaries from source, instead there is a `npm run fetch-invest` step that will look for pre-built binaries on GCS to download and bundle with the electron distribution.

`package.json` now has an "invest" property to configure which invest binaries will be fetched for workbench builds. 

The flask app that was formerly `server.py` will now live at `natcap.invest.web_api` and can be launched from the invest CLI (`invest serve`). That means the workbench only needs to include that one python executeable (the invest cli) when it formerly needed two. Those changes are evident in `main_helpers.js`.

This PR also drops binary builds for ubuntu - since we don't make invest binaries for ubuntu. And it skips the puppeteer test on macos because they reliably don't work at the moment (see #52 ).


davemfish:
Oh and there's a related invest PR that it probably makes sense to review kind of in parallel. https://github.com/natcap/invest/pull/399

Probably neither should be merged until we're happy with both being merged.

package.json
```diff
@@ -6,13 +6,14 @@
   "main": "src/main.js",
   "homepage": "./",
   "scripts": {
-    "start": "npx electron . --dev",
```
dcdenu4, line 9: Was the reason for dropping `npx` just because it's not really needed?

readme.md
```diff
@@ -9,29 +9,34 @@ workflows of an InVEST user.
...
-* `npm start`  
+* bind to an `invest` executeable (see package.json "invest" for a compatible version)
 
-(* invest-flask.spec script assumes a conda environment)
+In production, the invest exe comes from prebuilt binaries that are an artifact of the `invest` build process.  
 
-## To build this application
+For development, choose either:  
+  **A.** Duplicate the production setup by fetching prebuilt binaries `npm run fetch-invest`  
+  **B.** Use an any other locally installed, compatible, invest CLI (e.g. from a local python environment). To configure this, see `.env-example`
```
dcdenu4, line 18: Typo _Use an any_

readme.md
```diff
@@ -9,29 +9,34 @@ workflows of an InVEST user.
 ## To develop and launch this Application
 
 * `npm install`  
-* clone natcap.invest and checkout a recent revision (e.g. `main`)  
-* setup a conda* environment with deps for `natcap.invest Flask PyInstaller`  
-* build invest binaries  
-	`python -m PyInstaller --workpath build/pyi-build --clean --distpath build ./invest-flask.spec`  
-* `npm start`  
+* bind to an `invest` executeable (see package.json "invest" for a compatible version)
```
dcdenu4, line 12: Related to this and `invest` in `package.json`, is this something that developers are expected to change frequently and commit in PRs?

readme.md
```diff
@@ -9,29 +9,34 @@ workflows of an InVEST user.
...
+
+### To run various scripts and local programs
+See the "scripts" section of `package.json` and run them like:  
 `npm run lint`  
 `npm run test`  
 
-To run these or other command-line utils of locally installed packages outside the context of the `package.json scripts`, use `npx` (e.g. `npx eslint ...`) as a shortcut to the executeable. 
+To run other scripts or CLIs of locally installed packages, 
+prefix commands with `npx` (e.g. `npx eslint ...`). Otherwise, only
+globally installed packages are on the PATH.
```
dcdenu4, line 36: Ah, thank you.

dcdenu4 (commented):
Hey @davemfish This seems like a good move to me overall and I only had a few comments / questions. My main question / concern was around the `invest` setup in `package.json` and how this will work in practice for development and PRs. Should we also be providing a link those archives in the Readme from which to grab InVEST binaries from?

I'll add more thoughts on the InVEST side of things!

package.json
davemfish: Yeah, `npx` was pretty redundant inside this "scripts" context. I mainly use `npx` because it searches the local node_modules first for CLIs like electron, etc. But the "scripts" context already ensures that, I've since learned.

readme.md
davemfish: I don't think it will be changed frequenty, but yes, the "invest" property will need to be updated to whatever version of invest we want packaged with the workbench. Maybe in the future we would want to programmaticaly default to the latest invest release, but for now hardcoding the invest version in package.json seems okay to me.

davemfish:
This PR now also fixes #73

dcdenu4 (approved):
This looks good to me @davemfish !

emlys (approved):

emlys:
Looks like the mac test is failing to download the binaries, maybe a wrong url?

davemfish:
> Looks like the mac test is failing to download the binaries, maybe a wrong url?

Right, this should be due to the fact that the invest Mac builds failed on the invest commit this build is linked to.

phargogh (approved):
Looks good to me!

PR #79, opened 2021-01-12
Refactor UI spec "order" property

emlys:
As discussed in #31, this PR refactors the UI spec "order" property into its own top-level attribute.
- As it is now, the UI spec is no longer optional. Allowing it to stay optional would complicate this PR slightly so let's decide if that's a feature we still want.
- I have only edited one UI spec file (carbon) to fit the proposed new format. Once the change of format gets approved I'll update the other spec files.


src/components/SetupTab/index.jsx
```diff
@@ -83,18 +82,22 @@ function initializeArgValues(argsSpec, argsDict) {
...
+    // map each arg to an empty object, to fill in later
+    const argsValidation = Object.keys(props.argsSpec).reduce((acc, argkey) => {
+      acc[argkey] = {};
+      return acc;
+    }, {});
     this.state = {
       argsValues: null,
-      argsValidation: {},
-      argsValid: false,
-      sortedArgKeys: null,
```
emlys, line 97: since this is now just `this.props.uiSpec.order` I don't think it needs to be part of the `state` anymore

src/ui_data/natcap.invest.carbon.json
```diff
@@ -1,65 +1,56 @@
...
+        ["do_redd", "lulc_redd_path"],
+        ["do_valuation", "price_per_metric_ton_of_c", "discount_rate", "rate_change"]
+    ],
+    "args": {
+        "workspace_dir": {},
+        "results_suffix": {},
+        "lulc_cur_path": {},
+        "calc_sequestration": {
+            "ui_option": "disable",
+            "ui_control": [
```
davemfish, line 15: Even though I came up with them, these property names kinda trip me up, like I always have to pause and think about what they really mean. 

`ui_control` might be better as `control_targets` because it holds an array of the args that are _under control_ of `calc_sequestration`. And `ui_option` again is the option applied to the targets, not to the controller, so maybe better as `control_option`? Those are just suggestions, feel free to go with whatever makes sense to you.

davemfish (changes requested):
This looks like a great start @emlys . It looks like the code & the spec files will all be getting simpler and easier to read, so that's great! 

I had one minor comment on the spec itself. And then my other suggestion is to get the tests passing and then we can do another closer review of the changes from there. There are several tests that define UI specs and test the various functionalities. Ideally, those tests can be made passing just by updating their spec data.

src/ui_data/natcap.invest.carbon.json
emlys: `control_targets` and `control_option` sound good to me!

emlys:
After working on this more, I'm thinking that the UI spec files should be required, but each arg in the spec file should be optional. Since the `order` property is moved to the top-level, many args will have no properties. So this should be allowed (no `argsOptions.a`):
```
{
    order: [['a', 'b'], ['c']],
    argsOptions: {
        b: {control_targets: ['c']},
        c: {control_option: 'disable'}
    }
}
```

src/ui_data/README.md
```diff
@@ -0,0 +1,21 @@
+# Explanation of UI spec JSON format
```
davemfish, line 1: Nice addition!

src/ui_data/README.md
```diff
@@ -0,0 +1,21 @@
...
+* `"argsOptions"`: A mapping from arg keys to configuration options for displaying the corresponding input fields.
+    Each arg key is optional; if no configuration options apply, you can leave it out of this mapping.
+    Currently two configuration options are recognized:
+
+        * `"control_targets"`: A list of arg keys whose rendering is affected by the state of this arg key.
+            Each arg key in this list must also exist in `"argsOptions"` and have a `"control_option"` property.
+            
+        * `"control_option"`: A string that determines the CSS style to apply to this arg's input field, conditional
+            on the state of the controlling field. If `"control_option": "x"`, then the style `arg-x` is applied.
+            Currently `"disable"`, `"hide"`, and `"group"` are available.
```
davemfish, line 21: I don't think "group" is an option, is it?

tests/app.test.js
```diff
@@ -375,6 +376,15 @@ describe('InVEST subprocess testing', () => {
...
 
+  const uiSpec = {
+    order: [['workspace_dir', 'results_suffix']],
+    argsOptions: {}
+  }
+  const uiSpecFilePath = path.join(
+    fileRegistry.INVEST_UI_DATA, `${spec.module}.json`
+  );
+  fs.writeFileSync(uiSpecFilePath, JSON.stringify(uiSpec));
+
```
davemfish, line 387: Will these files get cleaned up after the tests? We might want a `beforeAll` and `afterAll` functions to take care of that.

tests/setuptab.test.js
```diff
@@ -340,18 +349,27 @@ describe('UI spec functionality', () => {
       args: {
         controller: {
           name: 'afoo',
-          type: 'csv',
-          ui_control: ['arg2'],
+          type: 'csv'
```
davemfish, line 352: A few lines above where I can't leave a comment, there's a code comment that I think we can get rid of now.

davemfish (changes requested):
I really like this redesign! I left some miscellaneous comments inline to check out. And we should look into the failing puppeteer test. I can try to reproduce that on Windows if it's not easy for you to reproduce and debug.

And one more thought I had is about some sort of testing/validation for the json files since they are now required and things might break in unexpected ways if one is missing or malformed. Do you have any thoughts about this? I could imagine a CI step that just validates their structure without running any part of the React app. Or way at the other end of the spectrum, we could have puppeteer click on every model to see that it loads. Whatever we do I think it's okay to push it to a separate issue, maybe flagged with the beta release milestone.

src/components/SetupTab/index.jsx
```diff
@@ -250,19 +210,17 @@ export default class SetupTab extends React.Component {
     *
     * @params {object} argsDict - key: value pairs of InVEST arguments.
     */
-    const { argsSpec } = this.props;
-    let { argsValues, argsValidation } = initializeArgValues(argsSpec, argsDict);
-    Object.keys(argsSpec).forEach((argkey) => {
+    let { argsValues } = initializeArgValues(this.props.argsSpec, this.props.uiSpec, argsDict);
+    Object.keys(this.props.argsSpec).forEach((argkey) => {
```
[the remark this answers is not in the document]
emlys: Ah, it should be `uiSpec.order`. The only difference is that `uiSpec.order` only includes the args that are supposed to show up in the form, so no `n_workers`, or `port` or `hostname` for the rec model.

src/ui_data/README.md
emlys: There is a `.arg-group` class in styles.css, but I don't think we have a use for it at the moment. I'll take it out of the README

tests/app.test.js
emlys: Moved the `writeFileSync` into `beforeAll` and added `unlinkSync` to `afterAll`

emlys:
Thanks for catching those miscellaneous things @davemfish. I agree some kind of validation would be good. Since a lot is changing in https://github.com/natcap/invest-workbench/pull/84 as well I'd wait until both PRs are done. In general, I like your idea of clicking on each model to make sure it loads.

emlys:
The failing windows test error looks like this one: https://github.com/electron/electron/issues/25267

emlys:
Closing because all changes have been merged into https://github.com/natcap/invest-workbench/pull/84.

PR #85, opened 2021-01-20
Bugfix/82 invest logging markup

### LINKED ISSUE #82: logfile markup is not working with invest 3.9.0 logfiles
We slightly changed the format of logging messages in invest to include `name` in parens like: `2021-01-15 07:14:37,148 (natcap.invest.hydropower.hydropower_water_yield) hydropower_water_yield.execute(268) INFO Validating arguments`

We'll need to update the `LogTab` markup function accordingly.

davemfish:
Fixes #82 

This PR adds test coverage for the log component's rendering and markup of invest logs. It's compatible with invest 3.9.0 logs, where we slightly altered the format of the log messages.

### DIFF
--- a/tests/logtab.test.js
+++ b/tests/logtab.test.js
@@ -2,12 +2,27 @@ import path from 'path';
 import fs from 'fs';
 import React from 'react';
 import {
-  fireEvent, render, waitFor, within
+  render, waitFor, within
 } from '@testing-library/react';
 import '@testing-library/jest-dom';
 
 import LogTab from '../src/components/LogTab';
 import { cleanupDir } from '../src/utils';
+
+function renderLogTab(logfilePath, primaryPythonLogger) {
+  const { ...utils } = render(
+    <LogTab
+      jobStatus="success"
+      logfile={logfilePath}
+      logStdErr=""
+      procID={undefined}
+      pyModuleName={primaryPythonLogger}
+      terminateInvestProcess={() => {}}
+      sidebarFooterElementId="divID"
+    />
+  );
+  return utils;
+}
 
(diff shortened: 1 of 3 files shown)


tests/logtab.test.js
```diff
@@ -0,0 +1,143 @@
...
+    cleanupDir(workspace);
+  });
+
+  test('Text in logfile is rendered', async () => {
+    const { findByText } = renderLogTab(
+      logfilePath, primaryPythonLogger
+    );
+
+    const log = await findByText(new RegExp(uniqueText));
+    expect(log).toBeInTheDocument();
```
dcdenu4, line 85: Hey Dave, what's this do `toBeInTheDocument()`?

dcdenu4: Specifically, what's the document?

dcdenu4 (approved):
Looks good to me! I certainly will have to start learning more about testing in this framework.

tests/logtab.test.js
davemfish: @dcdenu4 That'd be the "D" in DOM. i.e. the html page that is rendered by react. These `findBy` queries are querying the DOM and returning a specific node. Then, we're just expecting the node `toBeInTheDocument`. In this case, `findByText` would actually have raised an error if it couldn't find the text, so the `expect` is a little extraneous. But it would be weird to leave it out, I think.

The queries are from `testing-library` 
https://testing-library.com/docs/queries/about/

The "matchers" (`expect().toBe...`) are from jest or jest-dom
https://jestjs.io/docs/en/expect
https://github.com/testing-library/jest-dom

Issue #40, opened 2020-08-07
Log the user's GDAL_DATA path if it is present


davemfish:
Sometimes user's have a `GDAL_DATA` env variable on their system that causes conflict with the gdal libraries in invest. It's useful to know if that is in fact happening, so logging the GDAL_DATA variable should help.

davemfish:
closed by #88

PR #88, opened 2021-01-22
Window Menubar customization

### LINKED ISSUE #32: Window Menu Bar customization
The Electron defaults are okay for some menus (File, Edit, View, Window) but not okay for others (Help). We should customize these as needed so users get what they would reasonably expect.

davemfish:
This PR fixes #32 and #40 

It adds a custom (though mostly default) Menu bar to the top of the main electron window. The `About` menu now has an "About" page (mostly copied from invest's About content) and a "Report a problem" page with instructions on how to report problems and find the workbench logfiles.

When we release this to a wider audience, we should probably either drop the Report page altogether or streamline the reporting process. So this is kind of a quick-and-dirty solution just for the alpha-testers. Also, when you click this menu, node immediately logs it's entire `process` variable (which is why this should cover issue #40).

During this PR I learned that it's not okay store booleans in `process.env`, so there are some changes related to storing and passing around a flag for whether we're in dev-mode or not.

Finally, there's no testing here because I honestly don't know how to do it. These menus are outside the bounds of the DOM and I didn't find a puppeteer interface to them either. Also, these are mostly static pages, with just some very small bits of javascript.


phargogh:
@davemfish I see this PR is marked as a draft, so I'll hold off for now, but very happy to take a look whenever you're ready!

davemfish:
> @davemfish I see this PR is marked as a draft, so I'll hold off for now, but very happy to take a look whenever you're ready!

Whoops, forgot to undraft this, thanks for the reminder!

src/static/report_a_problem.html
```diff
@@ -0,0 +1,50 @@
...
+  <script>
+    const { shell, ipcRenderer } = require('electron');
+    const { getLogger } = require('../logger');
+    const logger = getLogger(__filename.split('/').slice(-1)[0]);
+    ipcRenderer.on('variable-reply', (event, arg) => {
+      userDataPath = arg.userDataPath;
+    });
+    ipcRenderer.send('variable-request', 'ping');
+    logger.debug('PROBLEM REPORT: process dump');
+    logger.debug(JSON.stringify(process));
```
phargogh, line 40: There sure is a lot of information written from this `JSON.stringify` call!  Would it be worth pretty-printing this, if there's an easy way to do it?

src/static/about.html
```diff
@@ -0,0 +1,177 @@
...
+          <td>pyamg</td>
+          <td>BSD</td>
+          <td>
+            <a
+              href='http://github.com/pyamg/pyamg'
+              onClick="handleClick(event)">
+              http://github.com/pyamg/pyamg
+            </a>
+          </td>
+        </tr>
```
phargogh, line 97: Ha, this is ancient!  `pyamg` is no longer shipped with InVEST, so we should pull this out of the other InVEST pane, too.

src/static/about.html
```diff
@@ -0,0 +1,177 @@
...
+          <td>PyQt</td>
+          <td>GPL</td>
+          <td>
+            <a
+              href='https://riverbankcomputing.com/software/pyqt/intro'
+              onClick="handleClick(event)">
+              https://riverbankcomputing.com/software/pyqt/intro
+            </a>
+          </td>
+       </tr>
```
phargogh, line 119: Hm I suppose this is out of date as well, since we  now use PySide2.  Uffda ... we should probably look back into licensing documentation issues at some point.

phargogh (changes requested):
Hey Dave, this is great!  The menu items I'd expect to show up on my mac are showing up as expected, and the "Find my workbench logs" button is a great idea.  I left a few minor comments throughout, but the main thing that doesn't quite have a line to comment on is that when I open the "About" and "Report a Problem" dialogs, I don't have a visual way to close them.  I can still use the dev tools to type `close()` and that works great, but otherwise the only way to return to the main window is to close the application and relaunch.

src/static/about.html
davemfish: Come to think of it, maybe we can remove this one altogether since we'll ultimately ship the workbench without PyQt or PySide2.

davemfish:
Hey @phargogh , I took care of all the little one-off items here. But I didn't have any luck figuring out the close browser window options for Mac. To clarify, the traffic lights are missing on these child windows, but present on the main electron window? I might need your help trying out the different options https://www.electronjs.org/docs/api/browser-window#new-browserwindowoptions

From what I can tell, those buttons should not be hidden by default.

phargogh:
> To clarify, the traffic lights are missing on these child windows, but present on the main electron window?

Yep!  The traffic light buttons are missing.  Here's a screenshot of what I see in the dialog:

Screen Shot 2021-01-25 at 10 52 48 AM

I'll take a look at those options you linked.

phargogh:
@davemfish much to my surprise, the `modal: true` option was what was causing the issue!  When I use the default `modal: false` on the child `BrowserWindow`s, the stoplight buttons show up and work well.  Also to my surprise, having the dialog be non-modal did _not_ allow the user to interact with the parent window ... which is inconsistent with typical interpretations of modality.

Here's the diff:

```diff
diff --git a/src/menubar.js b/src/menubar.js
index 0703fdc..49e53c8 100644
--- a/src/menubar.js
+++ b/src/menubar.js
@@ -108,7 +108,6 @@ function menuTemplate(parentWindow, isDevMode) {
 function openAboutWindow(parentWindow, isDevMode) {
   const child = new BrowserWindow({
     parent: parentWindow,
-    modal: true,
     width: 700,
     height: 800,
     frame: true,
@@ -128,7 +127,6 @@ function openAboutWindow(parentWindow, isDevMode) {
 function openReportWindow(parentWindow, isDevMode) {
   const child = new BrowserWindow({
     parent: parentWindow,
-    modal: true,
     width: 700,
     height: 800,
     frame: true,
```

And here's what it looks like now:

Screen Shot 2021-01-26 at 11 32 14 AM

davemfish:
Thanks for testing that, James! I let modal be false, now. It does seem weird that `modal: false` prevents interaction with the parent window on Mac. On Windows I can still interact with parent window. But I think it's okay either way, that's much need to leave both windows open at the same time. Back to you, @phargogh

phargogh (approved):
Yep, pretty weird about the modality there, and looks good!

PR #89, opened 2021-01-27
Bugfix/86 log messages display delayed

### LINKED ISSUE #86: invest log messages do not display in real-time
I just noticed this on Windows (didn't seem to notice when developing on linux). The log component uses npm `tail` package (mimic of unix `tail`) to stream lines from the invest logfile on disk to the react component. The problem is, the lines tend to come in large chunks, rather than line-by-line, even when configured to "watch" or "follow" the file. No lines are missed, but the `Tail` doesn't seem to keep up with every change to the file.

I briefly tried npm's `better-tail`, but saw basically the same results.

If I use the actual unix `tail -f` to tail the same logfile in a shell at the same time, the node `tail` _does_ keep up and the log component updates with each new line as desired. So it seems like some OS-level filesystem events are behind this issue.

It may have been a bad decision to render the logfile by streaming from the file on disk. Another option is to listen to the stdout of the invest subprocess. The only challenge there is when re-loading a saved invest job, the stdout data is not available, so the invest log would still need to be read from the file on disk. But, at least in that case the file is not still being actively appended to.

So, I'm thinking a good solution is to 1) drop npm `tail` and `better-tail` dependencies. 2) Render "running" invest logs by listening to invest's stdout stream and 3) Render pre-existing "complete" invest logs - from a saved job - by reading in the whole file.

phargogh:
This PR fixes #86 by forcing `node-tail` to use `fs.watchFile` instead of `fs.watch` (the default).  It turns out that on Windows, the mechanism that `fs.watch` uses will only notify node when the watched file _closes_, which is a problem on long-running models as the logfile won't close until model completion.  The alternative, `fs.watchFile` will poll the watched file's modified time on an interval and then report back to node if it's changed.

The node docs suggest that `fs.watch` is more efficient, but it doesn't serve our needs in this case.  The proposed changes will only use `fs.watchFile` on Windows and on all other OSes use `fs.watch`, which is more efficient.

A few experiments exist in https://github.com/phargogh/experiment-windows-node-fs-watch if you'd like to try these out in person.

### DIFF
--- a/src/components/LogTab/index.jsx
+++ b/src/components/LogTab/index.jsx
@@ -119,6 +119,7 @@ export default class LogTab extends React.Component {
     try {
       this.tail = new Tail(logfile, {
         fromBeginning: true,
+        logger: logger,
       });
       let markup = Object.assign('', this.state.logdata);
       this.tail.on('line', (data) => {
--- a/src/components/LogTab/index.jsx
+++ b/src/components/LogTab/index.jsx
@@ -117,10 +117,34 @@ export default class LogTab extends React.Component {
 
   tailLogfile(logfile) {
     try {
-      this.tail = new Tail(logfile, {
-        fromBeginning: true,
-        logger: logger,
-      });
+      if (process.platform === 'win32') {
+        /* On Windows, node's `fs.watch` only reports back to node about the
+         * logfile changing when the file is closed.  Python's FileHandler,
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Thanks for digging into this, @phargogh . Looks like it works well now!

PR #90, opened 2021-01-28
Handle smaller window sizes

emlys:
* Put the tabs in a scrolling area
* Made more space for the tabs by:
  * Moved the `Load Parameters` and `Settings` buttons into a dropdown. I just learned that the three vertical dots button (⋮) is called a kebab menu! 🍡 
  * Removed the Home button. Now clicking on 'InVEST' in the top right takes you to the home tab. 
* Modified the sidebar and `Run` button so that they don't overlap in a small window

There's a lot more we could do in the future to make it more pretty and consistent with how other apps work. Some things that I considered:
- Items in the sidebar could go into a hamburger menu (≡) which is commonly used for navigation
- Instead of scrolling, tabs that don't fit could collapse into an ellipsis menu (…) which typically means "more options...". This seems to be more common when there's a constant set of options (e.g. the 'Code', 'Issues', 'Pull Requests', etc. tabs on https://github.com/natcap/invest-workbench/). Scrolling seems more common for tabs that the user makes (web browser tabs, text editor open files) The ellipsis menu could also hold the 'Settings' and 'Load Parameters' buttons.
- Settings could get its own icon button (⚙)

Also including a small quality-of-life thing to handle the error you get when the invest binaries can't be found.
Before:
```
$ npm start

> invest-workbench@0.1.0-alpha start /Users/emily/invest-workbench
> electron . --dev

2021-01-29T19:10:51.194Z [main_helpers.js] error: ENOENT: no such file or directory, access 'build/invest/invest'
(node:52498) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, access 'build/invest/invest'
    at Object.accessSync (fs.js:208:3)
    at Object.accessSync (electron/js2c/asar.js:453:38)
    at /Users/emily/invest-workbench/src/main_helpers.js:41:10
    at new Promise (<anonymous>)
    at findInvestBinaries (/Users/emily/invest-workbench/src/main_helpers.js:15:10)
    at App.createWindow (/Users/emily/invest-workbench/src/main.js:36:27)
    at App.emit (events.js:310:20)
(node:52498) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:52498) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

After: 
```
$ npm start

> invest-workbench@0.1.0-alpha start /Users/emily/invest-workbench
> electron . --dev

2021-01-29T19:08:56.203Z [main_helpers.js] error: ENOENT: no such file or directory, access 'build/invest/invest'
ENOENT: no such file or directory, access 'build/invest/invest'
InVEST binaries are probably missing.
2021-01-29T19:08:56.205Z [main_helpers.js] error: no existing invest installations found
2021-01-29T19:08:56.524Z [server_requests.js] debug: retry # 1
2021-01-29T19:08:56.827Z [server_requests.js] debug: retry # 2
```


dcdenu4:
Hey @emlys, I just wanted to provide some quick feedback after playing around with your changes before diving into code review. 
- I really like the `InVEST` as Home button.
- I like the kebab menu addition
- The QOL feedback from invest binary not found is great.
- When making the application smaller I'm getting a vertical scroll for the model tabs. Since the text for the model named tabs wraps on the screen becoming narrower a horizontal scroll doesn't appear unless very skinny. I think it'd be great if the model named tabs did not wrap their text such that no vertical scroll would be necessary. I think a horizontal scroll where the model tab names get truncated would be ideal... I'm not sure how difficult such an implementation would be. 

So before diving into a code review, I wanted to check if you think it's worth spending a little more time on that model tab scroll layout? Thoughts?

dcdenu4:
image

davemfish:
This looks like it's coming along great! Very smart to rename "Home" to "InVEST" and add horizontal scrolling to the model tabs.

I'm less excited about hiding the Settings & Load Parameters buttons behind the kebab menu, though certainly not opposed to it. It looks very clean & functional. Here's some other considerations:
* I doubt there's much of a use-case for opening more than about 2-3 model tabs at once. For sure we should keep it looking clean & functional for any number of model tabs, but I think the primary use-case we should design for is like 1 - 2 tabs open.
* "Settings" as a ⚙️ icon in the top-right makes a lot of sense since that's very conventional and saves on space
* I use the "Load Parameters" button _all the time_ . I don't know if users will do the same (I think we might need to rename it or add a tootip or something to help convey it's usefulness), but I worry if it's hidden behind the dots it will rarely get used. 

Just some opinions, thanks for hearing me out! @emlys & @dcdenu4

emlys:
@dcdenu4 I think I've addressed everything we talked about! I got rid of the vertical scroll on the tabs and the whole app body. To make the tab close buttons look better, I turned off text wrapping for the tab labels. In the future it would be nice to truncate them with an ellipsis to a certain min width before it starts to scroll. We could also make the X button only appear when you have that tab selected, like tabs in Firefox do. But this should be ok for now!

emlys:
@davemfish if you use the Load Parameters button a lot, I certainly don't want to hide it! I'll try to quickly draft a version with a ⚙️ Settings button and a Load Parameters button so we can compare them

emlys:
@davemfish @dcdenu4 I tried moving Settings into its own icon button and bringing back the Load Parameters button (https://github.com/emlys/invest-workbench/tree/task/78-v2). Let me know what you think! Was also considering renaming Load Parameters to Open since it's shorter and consistent with what 'Open' does in other apps.
image

dcdenu4:
> Was also considering renaming Load Parameters to Open since it's shorter and consistent with what 'Open' does in other apps.

I think that setup looks good. What about `Load Previous` or `Load Model`? Then the tooltip could say "load model from saved parameters" or something. 

Now that I'm looking at it. I also think eventually we should have a `Load from parameters` link under `Save to Python Script` that basically does what the `onDrop` function for a datastack / logfile does. If I'm not mistaken, `Load Parameters` opens a new tab for the model.

emlys:
> Now that I'm looking at it. I also think eventually we should have a Load from parameters link under Save to Python Script that basically does what the onDrop function for a datastack / logfile does. If I'm not mistaken, Load Parameters opens a new tab for the model.

As far as I can tell Load Parameters is similar to the onDrop function, it opens a file selector window to choose a datastack

dcdenu4:
> As far as I can tell Load Parameters is similar to the onDrop function, it opens a file selector window to choose a datastack

It looks like it creates a new `InvestJob` and calls `openInvestModel(job)`. But if I have a model open with those save options to the left, I feel like I expect a load option there as well to update that model. BUT, this is another issue not related to smaller window sizes! haha.

davemfish:
> > As far as I can tell Load Parameters is similar to the onDrop function, it opens a file selector window to choose a datastack
> 
> It looks like it creates a new `InvestJob` and calls `openInvestModel(job)`. But if I have a model open with those save options to the left, I feel like I expect a load option there as well to update that model. BUT, this is another issue not related to smaller window sizes! haha.

Clearly there's some UX improvements to made here as the behaviors aren't obvious! 

"Load Parameters" is like choosing a model plus dropping a datastack all-in-one. In other words, I use it from the "Home" tab to quickly load a sample datastack. I kinda like not having to choose the right model from the list first, but that's just me so far.

On the other hand, I agree that if there's already a model tab open, then the "Load" button might do something unexpected by opening a new model tab. For that reason, maybe "Open" is a better name because I'd be less tempted to "open" if I'm already on the model tab I want. And then it could make sense to have "Load" in the sidebar near the "Saves" that does do exactly what the drag-drop does.

emlys:
> "Load Parameters" is like choosing a model plus dropping a datastack all-in-one. In other words, I use it from the "Home" tab to quickly load a sample datastack.

Ohh okay. 
If you're on the Home tab:
* drag-and-dropping a file should open a new model tab and populate it
* selecting a file via the "Load Parameters/Open" button should open a new model tab and populate it

If you're on a model tab:
* drag-and-dropping a file:
  * if the file is for the model that's open: populate the model form
  * if the file is for a different model: error? or open and populate a new model tab?
* selecting a file via the "Load Parameters/Open" button should open a new model tab and populate it
  * maybe this button only belongs on the Home screen?
* "Load" button in sidebar:
  * if the file is for the model that's open: populate the model form
  * if the file is for a different model: error? or open and populate a new model tab?   

I feel like having both a "Load" button in the sidebar and an "Open" button at the top of a model tab is kind of redundant. Since you have to go back to Home to open a new model tab in the usual way, what if the "Open" option also lived on the Home page? 
We could also put "Open" and "Open in new tab" in the custom menu bar File menu, so you could still access it from anywhere.

davemfish:
Great, so now we have #98 for dealing with the load/open buttons placement. And I think we agree that the existing "Load Parameters" could be shortened to "Open" and/or moved to a spot on the InVEST tab. Maybe it makes sense to do whichever of those is useful for this PR's goal of making space on the top nav bar, and then we can revisit it during #98 ?

emlys:
Okay, I changed "Load Parameters" to "Open" and merged in the version with the settings gear icon button.

package.json
```diff
@@ -74,6 +74,7 @@
     "esm": "^3.2.25",
     "glob": "^7.1.6",
     "localforage": "^1.9.0",
+    "material-design-icons": "^3.0.1",
```
dcdenu4, line 77: We can revisit this, but I do wonder how much this will add to the build size?

src/components/SettingsModal/index.jsx
```diff
@@ -73,15 +73,30 @@ export default class SettingsModal extends React.Component {
 
     const nWorkersIsValid = validateNWorkers(this.state.localSettings.nWorkers)
 
+    // define a custom button component to have a gear icon and no backgroundp
```
dcdenu4, line 76: Ah! `backgroundp`

src/components/SettingsModal/index.jsx
```diff
@@ -73,15 +73,30 @@ export default class SettingsModal extends React.Component {
...
+    const CustomButton = React.forwardRef(({ children, onClick }, ref) => (
```
dcdenu4, line 77: I'd love to learn about the reasoning behind using `forwardRef` here. It's the first time I'm seeing it and just started reading through the React documentation.

dcdenu4 (changes requested):
Hey @emlys , this looks great. I just had a few comments and one small change request. I was curious about the `forwardRef` for the settings button. I'm not sure it'd be easier to chat about instead of writing about here.

package.json
emlys: I had thought it would be small, because the [docs](https://google.github.io/material-design-icons/) say that "The icon font weighs in at only 42KB in its smallest woff2 format and 56KB in standard woff format." But checking it with `du -sh node_modules/material-design-icons/` I get that it's 356 megabytes! Not sure if that necessarily correlates with the build size.

src/components/SettingsModal/index.jsx
emlys: My only reasoning was "I copied it from the bootstrap docs" lol. [Here](https://react-bootstrap.netlify.app/components/dropdowns/#custom-dropdown-components) is the example I used.

emlys: I think the bootstrap `Dropdown` uses it to control whatever you render in the custom element.

emlys:
@dcdenu4 thanks! the icons size may be a concern, but I kind of prefer that to having assorted SVGs floating around in our repo. This way we don't have to manage them at all, and it's clear that we didn't make them ourselves. If the build size is an issue we should revisit it though.

dcdenu4:
Hey @emlys I thought about the material-icon a little bit more and wonder if we should just use the library: https://material-ui.com/components/icons/ 

I think in this case we could just do:
`import SettingsIcon from '@material-ui/icons/Settings';`
`...`
```javascript
<SettingsIcon
  onClick={this.handleShow}
/>
```

Let me know what you think about this approach!

tests/app.test.js
```diff
@@ -150,7 +150,7 @@ describe('Various ways to open and close InVEST models', () => {
       findByText,
       findByTitle,
       findByRole,
-      findAllByText,
+      findAllByText
```
davemfish, line 153: FYI, eslint suggests using the trailing commas. I don't feel strongly, but if you see them, they're not there by accident!

dcdenu4 (commented):
Just officially marking as reviewed with the comment about the material icons.

src/main_helpers.js
```diff
@@ -47,6 +47,9 @@ export function findInvestBinaries(isDevMode) {
     }
     logger.info(`Found invest binaries ${investExe} for version ${investVersion}`);
     resolve([investExe, `${investVersion}`.trim(os.EOL)]);
+  }).catch(error => {
+    console.log(error.message);
+    console.log('InVEST binaries are probably missing.');
```
davemfish, line 52: I'm curious about the intent here. There's an earlier try/catch that does a similar thing (though maybe it's message could be improved). And the earlier `catch` also `throws`, which would now get caught here. If it's useful to catch that error, then maybe we simply don't want to `throw` it on line 46? Or are there other errors getting caught here that miss the earlier try/catch?

Also, I'd prefer a `logger.error` instead of `console.log`. Then it goes to the logfile and the console (when in dev mode).

src/components/SettingsModal/index.jsx
davemfish: This stuff is also new to me. I've avoided Refs because I don't fully understand when they're truly called for. 

Kinda related, I wonder about the possible approaches for using icons in our app in general. I don't know much yet, but it seems like there are some libraries that provide icons as react components directly, which might reduce the need for some of this boilerplate stuff. On the other hand, it might be nice to just go straight to the source, like you've done here. Anyway, it could be worth thinking about this a bit before committing one way or another.

Here's just one example https://material-ui.com/components/icons/#material-icons

davemfish (commented):
The UI looks great! I tried it out and everything is scrolling and moving very nicely! A couple more comments, the main one being how we want to manage using icons in general. Like should we use a library that provides react components directly?

davemfish:
> Hey @emlys I thought about the material-icon a little bit more and wonder if we should just use the library: https://material-ui.com/components/icons/
> 
> I think in this case we could just do:
> `import SettingsIcon from '@material-ui/icons/Settings';`
> `...`
> 
> ```js
> <SettingsIcon
>   onClick={this.handleShow}
> />
> ```
> 
> Let me know what you think about this approach!

Aha, beat me to it!

src/main_helpers.js
dcdenu4: This is a good catch Dave and comment about using `logger`.

emlys:
> Hey @emlys I thought about the material-icon a little bit more and wonder if we should just use the library: https://material-ui.com/components/icons/

@dcdenu4 @davemfish  I actually came across that library while researching scroll bars and it looks neat! Could be useful for improving UX stuff in the future too. I steered away from it at the time because there are bunches of neat-looking react component libraries out there. Choosing one and committing to its design/style felt way out of scope for this PR.

tests/app.test.js
emlys: Oh, good to know! I'll put it back.

src/main_helpers.js
emlys: The main intent here is to avoid the ugly `UnhandledPromiseRejectionWarning` which you currently get if something goes wrong in this function. Per [this stackoverflow post](https://stackoverflow.com/questions/40500490/what-is-an-unhandled-promise-rejection), every `Promise` is expected to have a `.catch()` that handles Promise rejection. So we're getting a Warning on top of an error, because we're not handling the error in any way. I hadn't noticed the earlier `try/catch` but I agree it doesn't make sense to have both. I don't think there's any need for the first one. 
Will replace `console.log` with `logger.error` :)

emlys:
If you like the look of the Material UI library, though, we can certainly try it out! I expect most component libraries would have something like the `SettingsIcon`, so if we do settle on one, it would be trivial to replace. I addressed the other comments in c9424e6.

dcdenu4 (approved):
Nice, thanks Emily! Looks good and seems to be working well.

PR #93, opened 2021-01-28
Task/91 fix react devtools & dependency updates

### LINKED ISSUE #91: Get react-devtools extension working
There's an unresolved issue with electron and electron-devtools-installer that is blocking the react-devtools extension from working within electron. https://github.com/electron/electron/issues/23662

A workaround seems to be running a standalone react-devtools server while developing the electron app.

davemfish:
This PR fixes #91 by adding support for developers to use a standalone react-devtools window during development.

I also did a bunch of minor/major version updates to dependencies including electron and react. Everything still works! Well except for the old way of integrating react-devtools via an extension, which remained broken!


davemfish:
> Hey @davemfish, thanks for making a work around for this. I clone your branch and got it working on my end. I did have to `npm install -g react-devtools` ( I installed globally because it seems convenient ). Not sure if this should be mentioned as necessary in the readme?

Great point. I too had it installed globally, but then also decided to include it in the project's devDependencies. But for the shell to find it from there requires invoking like `npx react-devtools`. I updated the readme.

I'm still working on that test error, I think.

davemfish:
Okay this is ready for you again, @dcdenu4 . A bit more content here to review after all.

readme.md
```diff
@@ -42,6 +49,6 @@ To run javascript outside the electron runtime, but with the same ECMAscript fea
 
 ## Client logfile locations:
 
-Windows: "C:\Users\dmf\AppData\Roaming\invest-workbench\"
-Mac: "~/Library/Application Support/invest-workbench/"
-Linux: "~/.config/invest-workbench/"
\ No newline at end of file
+Windows: "C:\Users\dmf\AppData\Roaming\invest-workbench\"  
+Mac: "\~/Library/Application Support/invest-workbench/"  
+Linux: "\~/.config/invest-workbench/"
```
dcdenu4, line 54: Looks like some whitespace snuck in!

src/components/InvestTab/index.jsx
```diff
@@ -233,7 +233,12 @@ export default class InvestTab extends React.Component {
         procID: null,
       }, () => {
         saveJob(job);
-        cleanupDir(tempDir);
+        fs.unlink(datastackPath, (err) => {
+          if (err) { logger.error(err); }
+          fs.rmdir(tempDir, (e) => {
+            if (e) { logger.error(e); }
+          });
+        });
```
dcdenu4, line 241: Did it make sense to remove this from `utils.js` because it wasn't used anywhere else?

dcdenu4 (changes requested):
Thanks @davemfish , just a pesky request and a question.

readme.md
davemfish: 😮 
two spaces is a newline in markdown-speak ;)

src/components/InvestTab/index.jsx
davemfish: It was used several places in tests, and this was the only instance in src. But that `cleanupDir` function was being flaky and had a bug that was intermittently causing tests to error. It didn't really feel like it was worth maintaining our own function.

davemfish:
> Thanks @davemfish , just a pesky request and a question.

Did I answer your question? Back to you!

readme.md
dcdenu4: Ha!

dcdenu4 (approved):

PR #95, opened 2021-01-29
switched mac build target from zip to dmg

### LINKED ISSUE #94: Cannot open mac binaries
To reproduce:
1. Download the `invest-workbench-macos-latest` artifact from https://github.com/natcap/invest-workbench/actions/runs/520960147 and unzip the file.
2.  Open (double-click or right-click > open). A message appears _"The application "InVEST Workbench" can't be opened"_. 

We are not signing the workbench binaries for mac. The message is different that what you'd get if the problem is due to a missing/expired/revoked certificate. Most likely something is wrong with the binaries or the app bundle structure.

davemfish:
Fixes #94 

We had problems with the zip, and the dmg just works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "mac": {
       "category": "public.app-category.business",
       "icon": "build/icon.png",
-      "target": "zip"
+      "target": "dmg"
     },
     "linux": {
       "target": [


davemfish:
I opened a separate issue for the failing test. https://github.com/natcap/invest-workbench/issues/96

emlys (approved):
Sweet!

PR #100, opened 2021-02-03
launch the flask subprocess with a logging level arg

### LINKED ISSUE #99: flask app should be spawned with an explicit logging level
We launch the flask app via the invest CLI, and by default that sets a logging level of `CRITICAL`, which filters out all the messages from `ui_server.py`. We can add command-line args to the `invest serve` command to set the logging level more appropriately. 

I think `DEBUG` is reasonable for now as things are under active development.

davemfish:
Fixes #99 

The invest CLI defaults to `CRITICAL`, but we have a bunch of useful `DEBUG` level messages in `ui_server.py` (and further python modules) that should make it to stdout.

### DIFF
--- a/src/main_helpers.js
+++ b/src/main_helpers.js
@@ -60,7 +60,7 @@ export function createPythonFlaskProcess(investExe) {
   if (investExe) {
     const pythonServerProcess = spawn(
       path.basename(investExe),
-      ['serve', '--port', process.env.PORT],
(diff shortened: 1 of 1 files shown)


phargogh (approved):
👍

phargogh:
Looks like there's still that one LogTab test failure, but that looks separate from this PR.  Merging!

PR #103, opened 2021-02-05
Fix tab switching on close

### LINKED ISSUE #101: closing a model tab can leave no tabs active
To reproduce, open multiple model tabs and close one of them. We're left with a blank pane as no tabs are active. The intent was to find the next/previous open tab and make that one active, falling back to the Home/InVEST tab when no other model tabs remain.

emlys:
Fixes #101 
There were two issues here:
* Referencing `job.navID` is a property, it needed to be `job.metadata.navID`
* Clicking the close button triggered the `Nav` `onSelect` handler as well as the `button` `onClick` function. The `onSelect` was switching back to a nonexistent tab. Fixed this by adding `event.stopPropagation()` to the button `onClick` function.
Added a pretty thorough regression test as well.


src/app.jsx
```diff
@@ -104,13 +104,14 @@ export default class App extends React.Component {
...
+      switchTo = openJobs[index].metadata.navID;
     } else if (openJobs[index - 1]) {
-      switchTo = openJobs[index - 1].navID;
+      switchTo = openJobs[index - 1].metadata.navID;
     }
+    this.switchTabs(switchTo);
     this.setState({
       openJobs: openJobs
-    }, () => this.switchTabs(switchTo));
+    });
```
emlys, line 114: Moved this out of the callback because it doesn't depend on the state being updated. unless i'm missing something important?

tests/app.test.js
```diff
@@ -634,3 +634,99 @@ describe('InVEST subprocess testing', () => {
...
+
+  afterAll(async () => {
+    jest.resetAllMocks();
+    jest.resetModules();
+  });
+
+  test('', async () => {
+    const { 
+      getByRole,
+      findByText,
```
phargogh, line 656: I think these two queries aren't currently being used ... maybe they were left over from from earlier development?

tests/app.test.js
```diff
@@ -634,3 +634,99 @@ describe('InVEST subprocess testing', () => {
     spy.mockRestore();
   });
 });
+describe('Tab closing and switching', () => {
```
phargogh, line 637: This test makes sense to me!

As a more general question, `eslint` is showing a bunch of messages on this file.  Are we sticking to the linting rules defined in `.eslintrc`?

phargogh (changes requested):
Makes sense to me!  I did have a minor comment or two on the test, but nothing significant.

src/app.jsx
davemfish: I agree also. The idea of switching in the callback was probably taken from a similar routine in `openInvestModel` (line 86) where it may be important to wait until the callback so that react has time to re-render with the newly opened model before trying to switch to it. But that's probably not an issue when closing a model.

tests/app.test.js
davemfish: Yes please! I encourage us to keep adding/removing from the `.eslintrc` rules to override defaults that we don't like, but I do think it's useful to comply with whatever rules we have in place.

davemfish: In those rules, I started using `warn` as opposed to `error` for rules that I sometimes like to follow, but sometimes like to use my own judgment and ignore.

tests/app.test.js
emlys: `getByRole` is used a couple times for getting the close button, but you're right `findByText` shouldn't be in there!

tests/app.test.js
emlys: Whoops, forgot to lint! Thanks for catching that.

phargogh (approved):

phargogh:
Thanks @emlys !

PR #104, opened 2021-02-10
Bugfix/96 unreliable windows tests

### LINKED ISSUE #96: unreliable tests related to the LogTab
There are a few tests that have been intermittently failing lately, maybe due to race conditions.

tests in `logtab.test.js` sometimes errors when cleaning up a directory that is expected to be empty, but isn't
https://github.com/natcap/invest-workbench/pull/95/checks?check_run_id=1794680552

And the Windows puppeteer test sometimes fails to find the expected invest log text. I can't find an example of that right now, but hopefully it will come up again while I work on this. It's possible both failures are related, but that's not confirmed yet.

davemfish:
This fixes #96 

The windows puppet test fail was hard to reproduce locally, but I think it was just prone to jest's default timeout period. 

The logtab tests were poorly organized, with shared variables and race conditions when the tests run in parallel, which is the default.

I was originally planning to add to the puppet test suite in this PR, but that turned into a much bigger project that will progress in a separate branch.


dcdenu4 (commented):
Hey @davemfish . Thanks for tagging me on this. Good to go through more tests and get familiar.

I get why the `beforeEach` / `afterEach` would have caused issues if running in parallel, but would it be possible to just make the logfile and render the LogTab straight up and pass those into the test functions? Would that still cause issues to have these running in parallel on the same rendered object, but not destory it until after all tests have passed?

davemfish:
> Hey @davemfish . Thanks for tagging me on this. Good to go through more tests and get familiar.
> 
> I get why the `beforeEach` / `afterEach` would have caused issues if running in parallel, but would it be possible to just make the logfile and render the LogTab straight up and pass those into the test functions? Would that still cause issues to have these running in parallel on the same rendered object, but not destory it until after all tests have passed?

Oh good question. Since these tests don't happen to include any user events, state updates, or re-renders then it probably would be okay for all the tests to share the rendered component and logfile. Or similarly, all the `expects` could have gone in one single `test()` scope. I think that would be fine but not necessarily better/worse. Having multiple `tests` arguably makes it easier to see which one is failing. And each test with it's own rendered DOM just feels safer even if it's not strictly necessary.

dcdenu4:
> Having multiple `tests` arguably makes it easier to see which one is failing. And each test with it's own rendered DOM just feels safer even if it's not strictly necessary.

Makes sense to me!

dcdenu4 (approved):

PR #106, opened 2021-02-11
Use electron-builder directly rather than actions workflow

### LINKED ISSUE #65: Drop the electron-builder Actions script
@emlys  pointed out some very valid concerns about using the samuelmeuli/action-electron-builder script for code signing:

> 1. Since it's not our repo, it's hypothetically possible that it could become insecure and expose the certificate and password
> 2. The built in code signing steps rely on us storing the certificate, base64 encoded, in github secrets. There's nothing wrong with this except it goes against our pattern of storing that stuff in Google Cloud

Looking at the actions script (https://github.com/samuelmeuli/action-electron-builder/blob/master/index.js), there's really not much to it, so here's an issue to just create our own script for these purposes.

emlys:
Fixes #65 
Taking these changes out of #67 since that's on hold indefinitely.

I'm leaving in the GCP deploy steps and setting `--publish never` for electron-builder. They have a [generic publish](https://www.electron.build/configuration/publish#genericserveroptions) option which we could probably use to publish to the google cloud bucket, but I don't see any advantage to switching over.

Also removed `ubuntu` from the test matrix.


.github/workflows/build-electron.yml
```diff
@@ -25,23 +29,33 @@ jobs:
...
-      if: ${{ always() }}
-      with:
-        build_script_name: build
-        args: -c.extraMetadata.main=build/main.js
-        github_token: ${{ secrets.github_token }}
-        # If the commit is tagged with a version (e.g. "v1.0.0"),
-        # release the app after building
-        release: ${{ startsWith(github.ref, 'refs/tags/v') }}
+    - name: NPM Install
+      run: npm install
```
davemfish, line 35: Looks like the npm install step is duplicated

davemfish (changes requested):
This is great @emlys . I tested out the Windows artifact - all good. I found a duplicated npm install step in the script, but other than that, all good.

I'm good with removing the ubuntu tests. We probably don't want to sink extra effort into maintaining them. Anecdotally, they have been the most reliable (or least sensitive) of the 3 OS - often passing when the others encounter some OS-specific issue. So that's probably why they stuck around so long after we stopped building the app for ubuntu.

.github/workflows/build-electron.yml
```diff
@@ -25,23 +29,33 @@ jobs:
...
+        # use the built version as the entrypoint
+        # make sure we don't publish to github releases
+        npx --no-install electron-builder build --mac -c.extraMetadata.main=build/main.js --publish never
+
+    - name: Run electron-builder on Windows
+      if: matrix.os == 'windows-latest'
+      env:
+        GH_TOKEN: env.GITHUB_TOKEN
+        DEBUG: electron-builder
+      run: npx --no-install electron-builder build --windows -c.extraMetadata.main=build/main.js --publish never
```
davemfish, line 55: I just remembered that we also have `npm run dist` and `npm run pack` in our package.json. I use `npm run dist` regularly on Windows (I never use `pack` and forget what the extra `--dir` flag does). Do you think it's useful to call `npm run dist` here? Or are the options we need in GHA different from what we would use locally? 

I think leaving out the `--windows` or `--mac` flags will default to building for the current platform. So it might be possible to consolidate these two OS-specific steps into to a single "Run electron-builder" step. But if you prefer being explicit with the OS flag, that's fine by me also!

davemfish (commented):
Just left one more comment to get your feedback on!

.github/workflows/build-electron.yml
emlys: whoops!

.github/workflows/build-electron.yml
emlys: ah, good idea! The `--dir` option is: "Build unpacked dir. Useful to test." I took `pack` out of `package.json` since we aren't using it, and added `--publish never` to the `dist` script. Also explicitly included the `build` command. I think leaving out the platform flag is fine too.

davemfish (approved):
Nice!

PR #108, opened 2021-02-11
Desktop icon / Splash screen / Tooltip

### LINKED ISSUE #105: InVEST icon is missing
We can track an icon file that electron-builder will use as the app's desktop icon.

davemfish:
This PR fixes #105 by adding an icon to this repo that is passed to electron builder. 

I also added a splash screen that displays our traditional invest logo until the main window is rendered. 

And I added a tooltip to the Open button since that button isn't really self-explanatory ( it wasn't any clearer with it's former name either).

Emily, do you mind downloading and trying the Mac installer to confirm the icon & splashscreen are working?


emlys:
@davemfish the icon is working for me but the splashscreen isn't

davemfish:
> @davemfish the icon is working for me but the splashscreen isn't. it looks like it's expecting there to be an `invest-logo.png` in `src/static`?

Forgot to add the file, thank you!

davemfish:
Looks like I might also need to debug the puppet test given the new addition of the splash screen BrowserWindow.

davemfish:
Okay @emlys thanks for your patience. All the checks are passing now. I did have to re-run them a couple times because a few of Mac integration tests were timeing out on macos. Unfortunately that's hard to pin down and reproduce so I guess if it continues to be a problem we'll make an issue for it.

emlys:
@davemfish I tried it out again and the splash screen is showing up! I'm not sure it's quite having the intended effect on mac, though. After the first time, it only takes about 1 second to load. The splash screen image is noticeably not there at first and shows up ~halfway through the loading time. Which again is really short so it's not a problem, but I would expect a splash screen to be there from the start.

davemfish:
> @davemfish I tried it out again and the splash screen is showing up! I'm not sure it's quite having the intended effect on mac, though. After the first time, it only takes about 1 second to load. The splash screen image is noticeably not there at first and shows up ~halfway through the loading time. Which again is really short so it's not a problem, but I would expect a splash screen to be there from the start.

Thanks, @emlys . This sounds more/less the same as on Windows: 
* Click the launcher 
* nothing happens for ~1s
* Splashscreen displays with blank main window behind it
* < 1s later,  main window is done loading and splashscreen disappears

The splashscreen & everything else loads with `app.on('ready')` event, so I assume the first second of nothingness is before that event fires. And I'm not sure there's any way to display anything before that event. 

https://www.electronjs.org/docs/api/app#event-will-finish-launching
Do you think it's worth trying that `will-finish-launching` just for MacOS?

emlys:
@davemfish I think it's fine like this! Maybe eventually we can look into speeding up the launch time overall.

src/main.js
```diff
@@ -74,15 +76,31 @@ const createWindow = async () => {
...
     },
   });
-
   const menubar = Menu.buildFromTemplate(
     menuTemplate(mainWindow, ELECTRON_DEV_MODE)
   );
   Menu.setApplicationMenu(menubar);
 
-  // and load the index.html of the app.
+  splashScreen = new BrowserWindow({
```
emlys, line 84: I wonder if moving this to the very start of this function would make it show up any faster?

src/main.js
```diff
@@ -74,15 +76,31 @@ const createWindow = async () => {
...
   mainWindow.loadURL(`file://${__dirname}/index.html`);
 
+  mainWindow.once('ready-to-show', () => {
+    splashScreen.destroy();
+    // We should be able to hide mainWindow until it's ready,
+    // but there's a bug where a window initialized with { show: false }
+    // will load with invisible elements until it's touched/resized, etc.
+    // https://github.com/electron/electron/issues/27353
+    // So for now, we have the splashScreen over a blank, white mainWindow.
+    // mainWindow.show();
```
emlys, line 101: should this still be commented out?

tests/loadbutton.test.js
```diff
@@ -0,0 +1,20 @@
...
+    <LoadButton
+      openInvestModel={() => {}}
+      batchUpdateArgs={() => {}}
+    />
+  );
+
+  const openButton = await findByRole('button', { name: 'Open' });
+  userEvent.hover(openButton);
+  const hoverText = 'Browse to a datastack (.json) or invest logfile (.txt)';
+  expect(await findByText(hoverText)).toBeInTheDocument();
```
emlys, line 19: maybe also `unhover` and assert that the tooltip goes away?

emlys (commented):
Thanks @davemfish! my only other comment is that one time I tried it, the splash screen stayed superimposed on the main window after it loaded. I can't reproduce it now so just something to be aware of in case it happens again.

davemfish:
> Thanks @davemfish! my only other comment is that one time I tried it, the splash screen stayed superimposed on the main window after it loaded. I can't reproduce it now so just something to be aware of in case it happens again.

Yikes. If that continues to be a problem then it's probably not worth having the splash screen at all.

src/main.js
davemfish: Great idea, I think that helped!

src/main.js
davemfish: Yes, but I updated the comment so that's no longer confusing.

tests/loadbutton.test.js
davemfish: good idea!

emlys (approved):
Thanks @davemfish!

PR #116, opened 2021-02-17
Reliable puppeteer test on MacOS in GHA

### LINKED ISSUE #52: MacOS puppeteer test failing

emlys:
Fixes #52 
I can't explain it but the problem is solved by waiting a full 20 seconds before trying to query the remote debugging port.

The test failure doesn't make sense in the first place because the only error is happening _in a try/catch_ and you know it's being caught because the `console.log`s in the `catch` are executed: [line 69](https://github.com/natcap/invest-workbench/runs/1981958733#step:10:69), [line 75](https://github.com/natcap/invest-workbench/runs/1981958733#step:10:75)

It says "Test suite failed to run" and it doesn't look like the actual test started. But the error is happening in the `afterAll`. How is that being called before the tests?

I guess calling `browser = await puppeteer.connect...` too early results in `browser` being `undefined`, hence the error `Cannot read property 'close' of undefined`. But shouldn't the `await` take care of that?


tests/binary_tests/puppet.test.js
```diff
@@ -57,22 +57,29 @@ function makeAOI() {
...
+
+  // get data about the remote debugging endpoint
   // so we don't make the next fetch too early
-  await new Promise((resolve) => setTimeout(resolve, 5000));
+  await new Promise(resolve => setTimeout(resolve, 20000));
   const res = await fetch(`http://localhost:${PORT}/json/version`);
   const data = JSON.parse(await res.text());
+
+  // connect to the debugging endpoint
   browser = await puppeteer.connect({
```
emlys, line 77: not sure I understand this, `browser` is connected to the debugger URL but it's also used in the test like the actual app?

davemfish: I'm not sure I understand it much better. `puppeteer.connect` is an alternative to `puppeteer.launch`, for starting the browser. A more typical chrome use-case would be `launch` chrome, init a new `page` and then goto a URL in that page and do stuff. But since we don't want to launch a plain chromium browser, we need to first launch our electron app, and then use puppeteer to connect to it, which it somehow does using the devtools debugging protocols.

davemfish (commented):
Thanks for keeping after this @emlys ! It will be really great to have these tests running again. @phargogh is also working on a PR that adds to these tests, so just in general it's great to be able to rely on them more.

I had one idea that might be worth trying if we haven't already - switching to `browserURL` instead of `browserWSEndpoint` in the `puppeteer.connect` options.

.github/workflows/build-electron.yml
```diff
@@ -51,7 +51,6 @@ jobs:
 
     - name: Test electron app with puppeteer
       uses: GabrielBB/xvfb-action@v1
```
davemfish, line 53: I'm pretty sure this xvfb is only used on linux systems, so we can probably drop this too.

phargogh: Yep, I just looked into this action and it just executes the command exactly as stated _unless_ it's on linux.  If on linux it'll use xvfb to simulate a display.

emlys: Oh, great!

emlys:
@davemfish sorry, I totally forgot about this one! Fixed the merge conflicts... was there anything else?

davemfish:
> @davemfish sorry, I totally forgot about this one! Fixed the merge conflicts... was there anything else?

No worries, @emlys . I know these tests changed a bit in another PR in the meantime. And now there's a problematic Windows-only filepath that needs fixing. Sorry that should haven't made it through to begin with!

```
// append to this filename and the image will be uploaded to github artifacts
// E.g. page.screenshot({ path: `${SCREENSHOT_PREFIX}screenshot.png` })
const SCREENSHOT_PREFIX = path.join(
  os.homedir(), 'AppData/Roaming/invest-workbench/invest-workbench-'
);
```

Maybe this screenshot stuff is too fragile to keep around. The idea was to save screenshots and the app's logfile to a known location and then have an Actions upload-artifacts step upload it. You can see how that works for windows 

```
- name: Test electron app with puppeteer
      run: npm run test-electron-app

    - name: Upload app logging from puppeteer to github
      uses: actions/upload-artifact@v2.2.1
      if: matrix.os != 'macos-latest' && always()
      with:
        name: 'puppeteer_log.zip'
        path: '~/AppData/Roaming/invest-workbench/invest-workbench*'
```

Screenshots aren't so important, but uploading the logfile could be really useful for debugging. Would you like to add an upload step for MacOS?

davemfish:
Oh and it looks like you probably also fixed #134 !

emlys:
@davemfish I like the screenshot step, seems like it could be really helpful! Modified that and the upload step for mac. And yeah this should fix #134 too.

src/logger.js
```diff
@@ -20,6 +20,7 @@ if (ipcRenderer) {
   userDataPath = app.getPath('userData');
   isDevMode = process.argv[2] === '--dev';
 }
+console.log('user data path:', userDataPath);
```
davemfish, line 23: Let's remove this before merging, if you don't mind.

davemfish (changes requested):
Thanks @emlys ! There's a `console.log` that we might as well remove before merging. Otherwise, glad to see these are passing!

src/logger.js
emlys: ugghh I removed it yesterday but forgot to push! ready to go now.

davemfish (approved):
Thank you!

PR #119, opened 2021-02-19
Add version & git revision identifiers to artifact filenames

### LINKED ISSUE #117: artifact filenames need more identifiers
I would find it useful if electron builder produced artifacts with names like we have for invest:

InVEST_3.9.0.post266+g02ed490a_x64_Setup.exe

Right now the workbench has filenames like: 

invest-workbench_win_0.1.0-alpha.exe

I like having the `postNNN` so that's it's easy to see which is the latest dev build. And the git revision is also useful to link it to a specific changeset.

For invest, it looks like python's `setuptools` probably handles this for us. In this project, I think we can get the extra data with git commands:

`git rev-list --count <latest tag>..HEAD` returns the number of commits after the latest tag/release.
`git rev-parse --short HEAD` gets the current revision

davemfish:
This fixes #117 

Now that we have our first tag ( `0.1.0-alpha`) we can use `git describe --tags` to get a string that brings more meaning to our artifact filenames. This is especially helpful for dev builds, where we can see how many commits after the tagged commit a build is at (6) as well as the changeset that was built:

> invest-workbench_win_0.1.0-alpha-6-g5e478f0.exe

And conveniently, `git describe --tags` when HEAD is a tag, makes a string without the revision hash. So that's nice for a release build:

> invest-workbench_win_0.1.0-alpha.exe

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -19,6 +19,11 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # fetch complete history
+
+    - name: Fetch git tags
+      run: git fetch origin +refs/tags/*:refs/tags/*
 
     - name: Install Node.js
       uses: actions/setup-node@v1
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -47,12 +47,12 @@ jobs:
     - name: Test electron app with puppeteer
       uses: GabrielBB/xvfb-action@v1
       if: matrix.os != 'macos-latest'
+      continue-on-error: true
       with:
         run: npm run test-electron-app
     
     - name: Upload installer artifacts to github
       uses: actions/upload-artifact@v2.2.1
-      if: ${{ always() }}
       with:
         name: invest-workbench-${{ matrix.os }}
         path: dist/invest-workbench_*
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Set variables for GCS deploy target
       if: github.event_name != 'pull_request'
       run: |
-        echo "VERSION"=$(git describe --tags) >> $GITHUB_ENV
+        echo "VERSION"=$(grep VERSION electron-builder.env | cut -d '=' -f2) >> $GITHUB_ENV
         echo "BUCKET=$([ ${{ github.repository_owner }} == 'natcap' ] \
           && echo 'gs://releases.naturalcapitalproject.org/invest-workbench' \
           || echo 'gs://natcap-dev-build-artifacts/invest-workbench/${{ github.repository_owner }}' \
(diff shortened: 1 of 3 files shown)


.github/workflows/build-electron.yml
```diff
@@ -61,10 +66,8 @@ jobs:
     # to authenticate for GCP and they are not accessible in pull request workflows.
     - name: Set variables for GCS deploy target
       if: github.event_name != 'pull_request'
-      # TODO: If this is a release, don't include the commit hash
       run: |
-        echo "PKG_VERSION=$(jq -r .version package.json)" >> $GITHUB_ENV
-        echo "GIT_REV=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+        echo "VERSION"=$(git describe --tags) >> $GITHUB_ENV
```
phargogh, line 70: Ah!  This is very neat ... I hadn't known about `$GITHUB_ENV`

.github/workflows/build-electron.yml
```diff
@@ -88,9 +91,9 @@ jobs:
...
         CLOUDSDK_PYTHON: ${{env.pythonLocation}}\python.exe
       run: |
-        gsutil -m rsync dist/ "${{ env.BUCKET }}/${{ env.PKG_VERSION }}_${{ env.GIT_REV }}/"
+        gsutil -m rsync dist/ "${{ env.BUCKET }}/${{ env.VERSION }}/"
 
     - name: Deploy artifacts to GCS - macOS
       if: github.event_name != 'pull_request' && matrix.os == 'macos-latest'
       run: |
-        gsutil -m rsync dist/ "${{ env.BUCKET }}/${{ env.PKG_VERSION }}_${{ env.GIT_REV }}/"
+        gsutil -m rsync dist/ "${{ env.BUCKET }}/${{ env.VERSION }}/"
```
phargogh, line 99: It looks like we have two places where `git describe --tags` is used to get the version, one here directly in the actions workflow YML, and once again in `build.js` when it's written to `electron-builder.env`.  Would it make sense to consolidate the source of truth for the version string?  Like, maybe what's here as `env.VERSION` can be drawn from `electron-builder.env` somehow?  It'd be frustrating down the road to forget to change the version string in one place but not the other and then we don't know where the binaries ended up on GCP.

phargogh (changes requested):
Very cool!  I just had one comment about maybe avoiding duplication of information, but what do you think?

.github/workflows/build-electron.yml
davemfish: Yes, good idea. I grepped for it. And I also renamed some variables to always call this string a "version" instead of a git rev since that's a little less accurate for what it is.

phargogh:
@davemfish would you like me to take another look at this?  I wouldn't want to merge if you were still working on something!

davemfish:
> @davemfish would you like me to take another look at this? I wouldn't want to merge if you were still working on something!

Yes, all yours! Sorry that wasn't clear. @phargogh

phargogh (approved):
Thanks @davemfish !  Looks great..

PR #120, opened 2021-02-25
Redirect second instance of the application to the first

### LINKED ISSUE #74: Direct subsequent Workbench launches to the already-open instance
The user should not be able to have more than one Workbench UI open at once. There does not seem to be a use case for why this would be useful and it causes issues with handling the Flask App server and requests. There is probably some good examples of how to do this redirect via the Google machine.

phargogh:
This PR ( fixes #74 ) adds some functionality to `main.js` to:
1. Construct a mutex on application start.  If we can't get the lock, another instance already has it and we exit.
2. In the odd case where an application is already open and someone launches the application from a terminal (notably on mac), the mutex still applies and we bring the original application into focus and un-minimize it.

I was able to create a test for the "exit if there's a second process" behavior, but I had a tough time trying to find a good way to spoof those top-level events.  @davemfish any suggestions?

PS: On my mac, there were a few extra packages that needed to be installed or else the application wouldn't launch.  I can revisit those if you like.


src/main.js
```diff
@@ -128,18 +128,40 @@ const createWindow = async () => {
...
-  // dock icon is clicked and there are no other windows open.
-  if (mainWindow === null) {
-    createWindow();
-  }
-});
+// Single instance lock so subsequent instances of the application redirect to
+// the already-open one.
+// Adapted from https://www.electronjs.org/docs/api/app#apprequestsingleinstancelock
+const gotTheLock = app.requestSingleInstanceLock();
+if (!gotTheLock) {
```
davemfish, line 135: Nice that they made this easy for us!

davemfish (commented):
> I was able to create a test for the "exit if there's a second process" behavior, but I had a tough time trying to find a good way to spoof those top-level events. @davemfish any suggestions?

Not sure if I know what you mean here. Was there more behavior you wanted to test but couldn't? Like the Mac 'second-instance' event? 

I guess the only way we're testing main process code right now is by actually running the packaged app with puppeteer. We could write unit tests for some of the units in `main.js` and `main_helpers.js` if we want to. Not sure that helps you here though.

src/main.js
phargogh: It really is!  Implementation-wise, it looks like it does ultimately come down to a lock file, which is [managed in the chromium source](https://github.com/electron/electron/blob/ddf3ef0a5fe948fdad4831e29110df25f50882be/chromium_src/chrome/browser/process_singleton_posix.cc#L723).  Although I haven't found the line of source that says so, the lock appears to be unique to our application, probably distinguished by application name in `package.json`.

package.json
```diff
@@ -69,12 +69,22 @@
   "author": "dmf",
   "license": "MIT",
   "dependencies": {
+    "@babel/helper-builder-react-jsx": "^7.12.13",
+    "@babel/helper-builder-react-jsx-experimental": "^7.12.11",
+    "@babel/helper-regex": "^7.10.5",
+    "@babel/plugin-transform-react-jsx-self": "^7.12.13",
+    "@babel/plugin-transform-react-jsx-source": "^7.12.13",
```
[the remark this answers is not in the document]
phargogh: All of the dependency issues turned out to be needed because I was using Node version 15 rather than the Node 14 LTS release.  Downgrading to Node 14 rendered these package updates unnecessary.  I've removed these added dependencies in 713e339.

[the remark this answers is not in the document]
davemfish: Thanks, James! Normally I'd say we should all update our Node instead of asking you to downgrade, but I see the LTS release is still on Node 14, and the latest electron is actually running Node 12. So for those reasons, and in the interest of keeping our dependencies list small, I like your solution :)

davemfish:
Hey @phargogh I  noticed one test was failing so I took the liberty to re-run the jobs. The failure is an ongoing, intermittent one and we have this issue for it. https://github.com/natcap/invest-workbench/issues/96

phargogh:
Thanks @davemfish !  I'm still a little concerned that I don't seem to be able to run the puppet tests on my workstation, but they appear to be working on GHA, so that's good.  I'll re-request your review.

phargogh:
Hang on ... actually, the puppet tests are failing, just not failing the actions job.  @davemfish shouldn't the puppet test failures mark the test run as a failure?

davemfish:
> Hang on ... actually, the puppet tests are failing, just not failing the actions job. @davemfish shouldn't the puppet test failures mark the test run as a failure?

Wow, good catch. I've never seen that happen before. GHA should absolutely not show that check passing.

phargogh:
>  GHA should absolutely not show that check passing.

OK great, thanks!  I agree.  I'm still fiddling with tests on my end to try to get things to work, but I'll  look into the tests-not-failing-the-build issue as well.

Since we have a couple of active things going on, I'll mark this PR as a draft and keep at it.

.github/workflows/build-electron.yml
```diff
@@ -50,14 +50,12 @@ jobs:
       run: npm run test-flask-app
 
     - name: Test electron app with puppeteer
-      uses: GabrielBB/xvfb-action@v1
       if: matrix.os != 'macos-latest'
-      continue-on-error: true
```
davemfish, line 55: Whoops, I added this recently, but I didn't realize it would give the success indicator on error. I thought it would show the failure but continue anyway. Anyway, thanks for changing it back!

tests/binary_tests/puppet.test.js
```diff
@@ -156,3 +160,24 @@ test('Run a real invest model', async () => {
...
+  // Open another instance of the Workbench application.
+  // This should return quickly.  The test timeout is there in case the new i
+  // process hangs for some reason.
+  const otherElectronProcess = spawnSync(
+    `"${binaryPath}"`, [`--remote-debugging-port=${PORT}`],
+    { shell: true }
+  );
+
+  // When another instance is already open, we expect an exit code of 1.
+  expect(otherElectronProcess.status).toBe(1);
```
davemfish, line 182: I know the puppeteer test is not running on Mac on GHA (though it will again soon, @emlys figured that out in #116 ), but I'm curious if you tried this on your Mac and what happens? Does this second `spawnSync` trigger the 'second-instance' event? And if so, what actually happens to `otherElectronProcess`? Does it also exit with code 1?

davemfish (commented):
Hey @phargogh , this looks great. I just had one question about the test on MacOS. It might be good to go, but just wanted to confirm since we don't yet have that test running in GHA.

tests/binary_tests/puppet.test.js
phargogh: `spawnSync` starts the process and then waits for it to exit, returning a process object that we can then inspect (which we use for the exit code status assertion).  So by the time we reach this assertion, `otherElectronProcess` should absolutely have exited, and I'd expect the first electron process to still be hanging around until jest kills it at the end of the test.

On my mac, this executes as expected (and does appear reliable), but the other test hasn't been running reliably yet ... haven't figured out why.

.github/workflows/build-electron.yml
phargogh: Yeah, I thought that's what it'd be doing as well!  Alas, it silently continued

phargogh:
I think this is ready @davemfish !

tests/binary_tests/puppet.test.js
davemfish: Oh right, `spawnSync` is sync, so the fact that it gets to the assertion with an exit code is probably evidence that the 'second-instance' event did it's thing.

davemfish (approved):
thanks, James!

Issue #28, opened 2020-06-30
Add code-signing to the build process


davemfish:
I believe this can be configured with electron-builder.

emlys:
On hold until we get an official Apple signing ID.

emlys:
#67 only did Windows code signing. Some code signing setup for macos is in there, but electron-builder still isn't using the certificate for some reason.

emlys:
I have been experimenting with this again and may have hit a roadblock with the certificate.

IMO electron-builder is poorly configurable, what configuration there is is poorly documented, and the output is not helpful. Would seriously consider other options for building the binaries if there are any.

On windows, electron-builder calls `SignTool`, which provides some more helpful output:
```
SignTool Error: No certificates were found that met all the given criteria.


The following certificates were considered:
    Issued to: 3rd Party Mac Developer Installer: Stanford University (CQRZ4E7K9U)

    Issued by: Apple Worldwide Developer Relations Certification Authority

    Expires:   Tue May 10 21:01:02 2022

    SHA1 hash: A7ED95540FFD543C095983E4758AF0A72F88A763


After EKU filter, 0 certs were left.
After expiry filter, 0 certs were left.
After Private Key filter, 0 certs were left.
```
So the EKU (Extended Key Usage) is the problem. 

## EKU
EKU codes specify what the certificate may be used for. I checked the Key Usage and Extended Key Usage codes on our certificate as follows:
```
# convert p12 to pem, since the x509 command only accepts pem format
# you'll need to enter the certificate password
$ openssl pkcs12 -in ~/Downloads/20220510Expiry-macOSInstallerDistributionCert.p12 -passout pass:12345 -out cert.pem

# display key usage info
$ openssl x509 -in cert.pem -noout -ext keyUsage
X509v3 Key Usage: critical
    Digital Signature

# display extended key usage info
$ openssl x509 -in cert.pem -noout -ext extendedKeyUsage
X509v3 Extended Key Usage: critical
    1.2.840.113635.100.4.9
```
[...]

PR #125, opened 2021-03-02
Task/28 code signing

davemfish:
This PR addresses half of #28 . Windows ~~and Mac~~ binaries are signed by electron builder after setting the necessary env variables with our Stanford certificates, passwords, keychains, etc.

@emlys this basically just worked with the changes you had already drafted. I updated to our new certificate - the one that expires in 2024, and that's basically all I had to do.

I tested the Windows artifact and it was clearly signed. Would you check the Mac artifact please?

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -39,6 +39,8 @@ jobs:
     - name: Run the build script
       run: npm run build
 
+    # Steps that use github secrets do not run in pull requests.
+    # These include codesigning & GCS deploy-related things.
     - name: Set up Python for gsutil
       # gsutil requires a python, which is not included on Windows
       if: github.event_name != 'pull_request' && matrix.os == 'windows-latest'
@@ -55,7 +57,7 @@ jobs:
           service_account_key: ${{ secrets.GOOGLE_SERVICE_ACC_KEY }}
 
     - name: Code-signing setup for macOS
-      if: matrix.os == 'macos-latest'
+      if: github.event_name != 'pull_request' && matrix.os == 'macos-latest'
       shell: bash
       env:
         P12_FILE_PATH: ~/Downloads/stanford_cert.p12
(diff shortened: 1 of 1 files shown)


emlys:
@davemfish thanks for taking a stab at this! I downloaded the artifact from the latest actions run on your branch and checked it with `codesign -v` and it says it's not signed. in the electron-builder build step I don't see any output about code signing. it almost looks like it was truncated. 
looks like there's at least one step that shouldn't be running on pr, that's causing the pr build to fail.

davemfish:
Thanks for the feedback, @emlys . This is ready for review again. electron-builder is skipping the codesign on Mac because of the "identity" issue. So we should leave #28 open. But Windows signing is working. 
digi_sig


One puppeteer test is failing on Windows, which is nothing new, but pretty annoying. Maybe I'll dig into that again in a different PR.

.github/workflows/build-electron.yml
```diff
@@ -39,6 +39,55 @@ jobs:
...
+    - name: Code-signing setup for Windows
+      if: github.event_name != 'pull_request' && matrix.os == 'windows-latest'
+      env:
+        P12_FILE: Stanford-natcap-code-signing-cert-expires-2024-01-26.p12
+        CLOUDSDK_PYTHON: ${{env.pythonLocation}}\python.exe
+      run: |
+        gsutil cp gs://stanford_cert/$P12_FILE ~/$P12_FILE
+
+        # all these variables are used by electron-builder
+        echo "CSC_LINK=~/$P12_FILE" >> $GITHUB_ENV
```
emlys, line 87: I'd forgotten what the difference is between `env:` and `$GITHUB_ENV`. Variables in `env:` are just available in this step, and variables in the `$GITHUB_ENV` file are available to future steps. So this looks just right!

emlys (approved):
Thanks @davemfish! I'm glad this was more straightforward for windows at least. We can leave the mac codesigning stuff in there and come back to it another time.

Issue #45, opened 2020-09-17
consider not using electron's remote module


davemfish:
Electron has decided to deprecate it's `remote` module, but it will live on in "electron-userland" - meaning it will be maintained (or not) by volunteers going forward. 'remote' is a wrapper around IPC calls that allows communication between the `main` and `renderer` processes.

https://github.com/electron/electron/issues/21408

Some think there are security & performance concerns with `remote`
https://medium.com/@nornagon/electrons-remote-module-considered-harmful-70d69500f31

Other's are less concerned (2nd half of article)
https://medium.com/better-programming/2020-005-electron-apps-are-getting-faster-and-safer-3c045c39f61f

emlys:
I just read through that issue thread and wow, people have a lot of opinions on this one. If they do end up moving it to `electron-userland`, I would think we should switch to using `ipc` directly. How about putting this issue on hold until they finalize that decision?

davemfish:
Haha, yes indeed. 

I agree, using `ipc` directly seems like a good idea. There's been a couple cases where I used `ipc` instead of `remote` for new things as they came up, but haven't prioritized refactoring the existing `remote` uses. So yeah, I think this can go on hold, but I'd say we should not add any new `remote` code. Thanks for your input!

dcdenu4:
Interesting reads. I think I fell in the Joe Honton, `remote` isn't that harmful if used correctly, camp. But, it doesn't seem to matter! It has been deprecated as of v12 https://www.electronjs.org/releases/stable?version=12 and they make it pretty clear that the `remote` module replacement (https://github.com/electron/remote) shouldn't be used:

>⚠️ Warning! This module has many subtle pitfalls. There is almost always a better way to accomplish your task than using this module. For example, ipcRenderer.invoke can serve many common use cases.

PR #127, opened 2021-03-05
Add persistent global settings

### LINKED ISSUE #43: Store global settings in a persistent place
Settings such as `n_workers` and the python logging level are configured in the Settings dialog. Right now these changes are stored in memory and settings are reset to some default values when the application launches. It makes sense to store the settings somewhere on the filesystem (e.g. `electron.remote.app.getPath('userData')`) so that changes persist between sessions.

It's likely that more settings will be added in the future too.

dcdenu4:
Description
------------
This PR implements the load / save of a global settings file using: `app.getPath('userData')/global-settings.json` which saves the
`nWorkers` and `loggingLevel` settings.

The file is written when saving and loaded on `componentDidMount`.

A reset button feature was also added to the settings next to the save button. The reset button sets the `nWorkers` and `loggingLevel` back to hard coded defaults. The reset button does not also save the settings.

The node version was also bumped in the githbub actions workflows to use 14. This was mainly to get newer functionality related to the `fs` library.

Testing
-------
I was having some issues where I found that in `app.test.js`, resetting the mocks (`resetAllMocks`) in an `afterEach` call was causing `electron.remote.app.getPath`, defined in `__mocks__/electron.js`, to return `undefined`. Thus I reset the mock implementation manually in the necessary `beforeEach` blocks.

Fixes #43


davemfish:
Hey @dcdenu4 I know this is still under draft, but I couldn't help myself and I have a couple high-level comments that are better sooner than later.

First, we could consider using `localforage` as the persistent store. I'm not suggesting we _should_ do that instead of the json file you're using. But since we're already using `localforage`, I thought I'd mention it. I think a json file is a nice simple solution. You can look at `InvestJob.js` to see how we're using `localforage`. It might be overkill for these settings.

Second, the `remote` module is being deprecated in electron v12 (see #45 ). It's still in use in various places, but we've been trying not to add more uses of it. So another way to get at the `getPath('userData')` from the renderer is by requesting the data from the main process. That's already setup in `main.js`, so you can make a call like:

```
const { ipcRenderer } = require('electron');
let userDataPath;
ipcRenderer.on('variable-reply', (event, arg) => {
  userDataPath = arg.userDataPath;
});
ipcRenderer.send('variable-request', 'ping');
```

Sorry the original issue didn't mention this stuff! It probably pre-dated us figuring these things out!

dcdenu4:
Closing this for a newer version.

PR #129, opened 2021-03-05
Removing the server calls to determine if a vector might have points

### LINKED ISSUE #128: Remove DelineateIt interactivity changes based on whether a vector may contain points
In InVEST's `main` branch, we recently removed the conditional enabling of the point snapping container based on whether there _might_ be any points.  This caused some unexpected behavior in the UI and was ultimately removed (https://github.com/natcap/invest/issues/479, https://github.com/natcap/invest/pull/518).  This should also be reflected in the workbench.

phargogh:
This PR removes the `getVectorMayHavePoints` endpoint because we're no longer using it in DelineateIt, and this also simplifies the DelineateIt UI in the workbench where the `getVectorMayHavePoints` function was used.

Plus some bonus whitespace chomping.

Parallel InVEST PR over at https://github.com/natcap/invest/pull/520 for the flask functionality.
Fixes #128

### DIFF
--- a/src/server_requests.js
+++ b/src/server_requests.js
@@ -130,25 +130,6 @@ export function getVectorColumnNames(payload) {
 }
 
 /**
- * Get a boolean indicating whether a vector file may contain points.
- *
- * @param {string} payload - path to file
- * @returns {Promise} resolves boolean
- */
-export function getVectorMayHavePoints(payload) {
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Looks good, thanks, James!

PR #130, opened 2021-03-08
Bugfix/96 unreliable windows puppeteer

### LINKED ISSUE #96: unreliable tests related to the LogTab

davemfish:
This PR fixes #96. There were two sources of unreliability in the puppeteer tests that are fixed here:

1. `BROWSER.waitForTarget` avoids a race condition and an occasional undefined `page` that yielded errors like `cannot find property mainFrame of undefined` (or something like that).
2. connecting with `defaultViewport: null` avoids some unpredictable resizing that seemed to happen because of "viewport emulation" just before the run button `click`. That sometimes caused the mouse to miss the button and the test to hang until the `logTab` expect timed out. This solution was mentioned here: https://github.com/puppeteer/puppeteer/issues/1183

This PR also adds a screenshot to the test, which helped with debugging, and it uploads that image and the app's logfile to a github artifact, which was very helpful for debugging behavior that wouldn't reproduce locally. I left that stuff in because it could prove useful again and doesn't add too much of burden, I don't think.

And there's also some general cleanup of the test script.

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -57,8 +57,8 @@ jobs:
       uses: actions/upload-artifact@v2.2.1
       if: matrix.os != 'macos-latest' && always()
       with:
-        name: 'puppeteer_log.txt'
-        path: '~/AppData/Roaming/invest-workbench/invest-workbench-log*'
+        name: 'puppeteer_log.zip'
+        path: '~/AppData/Roaming/invest-workbench/invest-workbench*'
 
     - name: Upload installer artifacts to github
       uses: actions/upload-artifact@v2.2.1
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -58,7 +58,7 @@ jobs:
       if: matrix.os != 'macos-latest'
       with:
         name: 'puppeteer_log.txt'
-        path: '~/AppData/Roaming/invest-workbench-log*'
+        path: '~/AppData/Roaming/invest-workbench/invest-workbench-log*'
 
     - name: Upload installer artifacts to github
       uses: actions/upload-artifact@v2.2.1
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -53,6 +53,13 @@ jobs:
       if: matrix.os != 'macos-latest'
       run: npm run test-electron-app
 
+    - name: Upload app logging from puppeteer to github
+      uses: actions/upload-artifact@v2.2.1
+      if: matrix.os != 'macos-latest'
+      with:
+        name: 'puppeteer_log.txt'
+        path: '~/AppData/Roaming/invest-workbench-log*'
+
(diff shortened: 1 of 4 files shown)


.github/workflows/build-electron.yml
```diff
@@ -102,6 +102,13 @@ jobs:
       if: matrix.os != 'macos-latest'
       run: npm run test-electron-app
 
+    - name: Upload app logging from puppeteer to github
+      uses: actions/upload-artifact@v2.2.1
+      if: matrix.os != 'macos-latest' && always()
```
dcdenu4, line 107: What's the `always()` doing in this case? I read that it returns `true` even if cancelled.

tests/binary_tests/puppet.test.js
```diff
@@ -8,29 +9,31 @@ import { getDocument, queries, waitFor } from 'pptr-testing-library';
 
 jest.setTimeout(120000); // This test takes ~15 seconds, but longer in CI
 const PORT = 9009;
+const TMP_DIR = fs.mkdtempSync('tests/data/_');
```
dcdenu4, line 12: I noticed the switch in variable naming convention. Curious what made you go with all-cap underscores?

tests/binary_tests/puppet.test.js
```diff
@@ -61,120 +64,108 @@ function makeAOI() {
...
-  let page;
-  pages.forEach((p) => {
-    if (p.url().endsWith('index.html')) {
-      page = p;
-    }
-  });
+  const target = await BROWSER.waitForTarget(
+    (target) => target.url().endsWith('index.html')
+  );
+  const page = await target.page();
```
dcdenu4, line 111: Nice, this makes sense to me.

dcdenu4 (commented):
Sorry for the non-assigned review! I was just curious if this would end up helping on my end. I also had a few curiosity questions @davemfish , but that's about it. 

I'll let @emlys review and approve / merge, but I couldn't help ask some questions!

.github/workflows/build-electron.yml
davemfish: Making sure this step runs even if the test in the prior step failed, since that's when the logs are most important.

tests/binary_tests/puppet.test.js
davemfish: Trying to make it clearer which variables have global scope.

emlys (approved):
Thanks @davemfish, this is great! The code for finding the page makes more sense, and the added screenshot step is cool.

PR #133, opened 2021-03-16
Feature/77 configure electron builder for app updates

### LINKED ISSUE #77: Figure out how to push updates to users
It would be nice to allow workbench updates in the style of modern browser updates, where new files seem to be patched into an existing installation. As opposed to installing a whole separate instance of the application and uninstalling the older one.

The extra resources of the python binaries _might_ make this more complicated, or maybe there's really nothing special about that in this context.

davemfish:
This PR fixes #77. It configures electron builder so that we can meet the following requirements for installations & updates of the workbench.

1. user's should have full control over when they want to update `natcap.invest` - no surprise/silent updates
1. users (and especially devs) should easily be able to have multiple versions installed. I often want the latest official release installed and want to be able to install & test a dev build without clobbering my other installation.

Without doing anything fancy, we can achieve this by managing workbench updates the same way we manage invest updates now: install new versions to a new install directory. For example, the install path can vary by invest version string, which is referenced in `package.json invest.version`. That means workbench updates that don't include a different `natcap.invest` version will overwrite an existing workbench. If that's undesirable, we can always vary the install path by workbench version string as well.

This PR moves the electron-builder config out of package.json and into a script so that we can dynamically set variables that include version strings.


.github/workflows/build-electron.yml
```diff
@@ -114,12 +114,12 @@ jobs:
       if: ${{ always() }}
       with:
         name: invest-workbench-${{ matrix.os }}
-        path: dist/invest-workbench_*
+        path: dist/invest_*_workbench_*
```
phargogh, line 117: Oh, I'm so glad they added globbing

electron-builder-config.js
```diff
@@ -0,0 +1,63 @@
...
+// Uniquely identify the changeset we're building & packaging.
+const workbenchVersion = execFileSync('git', ['describe', '--tags'])
+  .toString().trim();
+// Need to pass this data to the GHA deploy step
+fs.writeFileSync('.workbench_version_string.env', workbenchVersion);
+
+// productName controls the install dirname & app name
+// We might want to remove the workbench version from this name
+const PRODUCT_NAME = `InVEST ${pkg.invest.version} Workbench ${workbenchVersion}`;
+const ARTIFACT_NAME = `invest_${pkg.invest.version}_workbench_${workbenchVersion}_${OS}_${ARCH}.${EXT}`;
```
phargogh, line 23: This is very exciting, being able to configure these in code!

electron-builder-config.js
```diff
@@ -0,0 +1,63 @@
+const fs = require('fs');
+const { execFileSync } = require('child_process');
+const pkg = require('./package');
+
+const OS = process.platform;
+const ARCH = process.arch;
+const EXT = OS === 'win32' ? 'exe' : 'dmg';
```
phargogh, line 7: Would this mean that the appImage on Linux (since there's a configuration defined below) also have the `.dmg` extension?

phargogh (commented):
Just a couple questions, mostly for my own information.  The only reason I haven't approved/merged is about the use of the extension and its applicability to linux build targets.

electron-builder-config.js
davemfish: Thanks for pointing this out! I opted to remove the linux section of the config as the rest of our build process (specifically the invest side) doesn't support linux right now.

Fun fact, one thing we lost when moving this config out of `package.json` were some built-in template variables for these strings. So now we have to make them ourselves, but no big deal.

davemfish:
Thanks, @phargogh , back to you! I wonder if you tried installing & then re-installing on Mac, just to see if that also works as expected? 

Testing that is a bit of a pain since we don't have multiple versions built yet. But if you wanted you could overwrite `pkg.invest.version` in the config script and then build the app locally and install it. And then follow that up by installing this one https://storage.googleapis.com/natcap-dev-build-artifacts/invest-workbench/davemfish/0.1.0-alpha-127-g78f3a40/invest_3.9.0.post235%2Bg296690d7_workbench_0.1.0-alpha-127-g78f3a40_darwin_x64.dmg

Then I'd expect you to have two distinct installations. 🤞

electron-builder-config.js
phargogh: Oh that's interesting about the template variables being available in `package.json` but not here.  The approach you have here makes sense to me.

phargogh:
Well, I'm trying to finish this up and test out these parallel installs, but I'm having issues with running the workbench _at all_ on my mac laptop, _including_ with the mac build from the latest GHA build on `main`.  I'll try this on my windows computer as soon as I can to see if it's reproducible there.  If not, I'll report a separate issue for this that we can follow up on.

phargogh:
Alright, there's definitely something weird going on here, where electron is reliably crashing with `EXC_BREAKPOINT (SIGTRAP)`.  Even when clearing out `node_modules`, reinstalling and `npm start`ing, I _was_ getting the same crash.  And now I'm actually getting a separate crash for a different reason.  Clearly I need to debug this some more.

@emlys could you try running @davemfish 's suggestion in https://github.com/natcap/invest-workbench/pull/133#issuecomment-801989513 to see if this works as expected with parallel installs?  Thanks in advance!

davemfish:
> Alright, there's definitely something weird going on here on my mac, where electron is reliably crashing with `EXC_BREAKPOINT (SIGTRAP)`. Even when clearing out `node_modules`, reinstalling and `npm start`ing, I _was_ getting the same crash. And now I'm actually getting a separate crash for a different reason. Clearly I need to debug this some more.
> 
> @emlys could you try running @davemfish 's suggestion in [#133 (comment)](https://github.com/natcap/invest-workbench/pull/133#issuecomment-801989513) to see if this works as expected with parallel installs? Thanks in advance!

Thanks for trying, James!

emlys:
surprisingly I am having the same problem! I tried versions from different builds on main going back over a month and they all crash. probably not related, but I get "download failed" 90% of the time I try to download an actions artifact now.

davemfish:
Hi @phargogh , it looks like builds are passing again on this PR. Could you test out multiple installations on Mac? To remind ourselves, the expected behavior here is that multiple installations will stand side-by-side if the invest version differs between them. But if the invest version is the same, then a second workbench installation will overwrite the earlier one. Here's how this could be tested:

> Testing that is a bit of a pain since we don't have multiple versions built yet. But if you wanted you could change `pkg.invest.version` in the config script and then build the app locally and install it. Easiest thing is to not change the actual package.json, just overwrite that `pkg.invest.version` variable in `electron-builder-config.js`. That will just fake the product name string, and that's what matters for overwriting or not. 

And then follow that up by installing this one 

https://github.com/natcap/invest-workbench/suites/2541793519/artifacts/55188033

phargogh:
@davemfish I just ran the two versions of the workbench side-by-side following the procedure you described without any _electron_ issues.  I can see the different version strings in the menu bar and the application lock isn't preventing launch, so that works great!

However, I'm also seeing a separate issue where the application that's causing the whole webpage to go blank, which I've posted as #142 , but I don't think this is related to electron.

phargogh (approved):
We decided to call this PR complete and work on the newly-discovered mac issue #142 separately

PR #135, opened 2021-03-17
Deprecate remote use and test on Electron 12

### LINKED ISSUE #45: consider not using electron's remote module

dcdenu4:
This PR deprecates the use of `remote` as it will no longer be supported by Electron going forward. See issue natcap/invest-workbench#45 for a recap of why the decision was made. 

I tested Electron v. 12 with this PR and had to make one change due to a breaking change where `contextIsolation` is now by default set to `true`. So the change was defaulting it to `false`.  https://www.electronjs.org/docs/tutorial/context-isolation
>Another implication is that require() cannot be used in the renderer process unless nodeIntegration is true and contextIsolation is false.

https://www.electronjs.org/docs/breaking-changes#default-changed-contextisolation-defaults-to-true

That seemed outside the scope for this PR, so I reverted to using Electron 11.3.0. Electron says:
> We recommend having contextIsolation enabled for the security of your application.

Fixes natcap/invest-workbench#45 

Just saw that issue natcap/invest#870 talks about the security issues.


src/main.js
```diff
@@ -59,9 +61,37 @@ const createWindow = async () => {
...
+  })
+  ipcMain.handle('show-open-dialog', async (event, arg) => {
+    const result = await dialog.showOpenDialog(arg);
+    return result;
+  });
+  ipcMain.handle('show-save-dialog', async (event, arg) => {
+    const result = await dialog.showSaveDialog(arg);
+    return result;
+  });
+  ipcMain.handle('is-dev-mode', async (event) => {
```
davemfish, line 87: Because we have this, we may no longer need the extra `webPreferences` when the `BrowserWindow` is instantiated. That was another "non-remote module" way to let the renderer know if we're in devmode. The root `index.js` looks for this flag.
```
additionalArguments: [
        ELECTRON_DEV_MODE ? '--dev' : 'packaged'
      ],
```
It could still be useful to pass the flag like this. Or we could add another IPC call instead. What do you think?

davemfish: (github won't let me comment on the relevant line)

tests/app.test.js
```diff
@@ -110,7 +110,7 @@ describe('Various ways to open and close InVEST models', () => {
       model_run_name: 'carbon',
       model_human_name: 'Carbon',
     };
-    remote.dialog.showOpenDialog.mockResolvedValue(mockDialogData);
+    ipcRenderer.invoke.mockResolvedValue(mockDialogData);
```
davemfish, line 113: Hmm this is interesting. I was wondering if it made sense to be a little more specific with the mock and mock the `electron.dialog.showOpenDialog` that is called in `main.js`. But none of this IPC actually works in tests does it, since the electron processes are not actually running.  

So this seems right to me as basically equivalent to the tests when we used `remote`. But also makes me wonder if it's worth trying to cover more of this IPC code with tests? It could be done with puppeteer, or I guess it could be done by mocking the implementations of `ipcRenderer` and `ipcMain` with actual `Event` objects. I'm not sure if the latter is really worth the effort.

davemfish (commented):
Hi @dcdenu4 , this looks great. No definite changes to request but a few questions or things to talk over before merging. Let me know what you think!

src/constants.js
```diff
@@ -1,7 +1,12 @@
...
-import { remote } from 'electron'; // eslint-disable-line import/no-extraneous-dependencies
+import { ipcRenderer } from 'electron';
 
-const USER_DATA = remote.app.getPath('userData');
+let USER_DATA = '';
+
+ipcRenderer.on('variable-reply', (event, arg) => {
+  USER_DATA = arg.userDataPath;
+})
+ipcRenderer.send('variable-request', 'ping');
```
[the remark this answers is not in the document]
dcdenu4: Great question. I looked into this slightly. I think this use case is here because `invoke` wasn't working and I never went back and switched it. I'll look into it more but my take away is:

`on / send` is always live and listening and will take action on any `variable-reply`.

`invoke` is more of a component requesting something once because this component or function was called to do something.

Probably in most our cases `invoke` makes sense. I can go back and look to see if there's any reason to keep some of the `on / send` calls.

src/components/LoadButton/index.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { remote } from 'electron'; // eslint-disable-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
dcdenu4: Yep, I see that too. I think I accidentally deleted some of these when switching to `ipcRenderer`.

src/index.js
```diff
@@ -19,21 +19,12 @@ const app = require('./app');
...
-  }
-}));
-
 window.addEventListener('contextmenu', (e) => {
   e.preventDefault();
   rightClickPosition = { x: e.x, y: e.y };
-  menu.popup({ window: remote.getCurrentWindow() });
-}, false);
+  ipcRenderer.send('show-context-menu', rightClickPosition);
+})
```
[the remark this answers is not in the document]
dcdenu4: Yea! It took me a hot second to figure out the right divide between parts, but definitely cleaner I think.

src/logger.js
```diff
@@ -1,14 +1,20 @@
...
+const { app, ipcRenderer } = require('electron');
 
-let userDataPath;
+let userDataPath = '';
 let isDevMode;
-if (remote) {
-  // When this module is imported from render process, access via remote
-  userDataPath = remote.app.getPath('userData');
-  isDevMode = remote.process.argv[2] === '--dev';
+if (ipcRenderer) {
```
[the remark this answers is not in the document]
dcdenu4: haha, yes. I was confused as well. I can confirm that I tried not using that and the tests failed. I can look more into this magic if you'd like. I was certainly assuming the developer knew something I didn't.

src/main.js
dcdenu4: Hmmm good question, I'll take a look.

tests/app.test.js
dcdenu4: Yea, when re working the tests it took me awhile to realize that this was just going to be pretty empty shell stuff for mocking `ipc`.

I think I agree that it might be better to cover this using puppeteer. I can try to add a few tests and see how it goes.

src/constants.js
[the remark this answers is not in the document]
dcdenu4: After a little more looking into this I think that `invoke / handler` is new syntax to better help with `promises` and using `async / await`. Though `send / on` could be useful if you had something updating another state like a countdown or progress bar or something.

src/main.js
dcdenu4: I'm fine leaving this in. Six in one hand, half dozen in the other.

tests/app.test.js
dcdenu4: We decided to punt on this for now.

dcdenu4:
@davemfish assuming that mac test passes, this is back to you! I did some linting and swapped out `send / on` with the new `handle / invoke` style syntax. Let me know what you think of how I included `catch` on some of the promise handling.

src/static/about.html
```diff
@@ -137,15 +137,22 @@ <h2>Open-Source Licenses</h2>
...
+      const text1 = document.createTextNode(response.investVersion)
       node1.appendChild(text1)
-      const text2 = document.createTextNode(arg.workbenchVersion)
+      const text2 = document.createTextNode(response.workbenchVersion)
       node2.appendChild(text2)
     })
-    ipcRenderer.send('variable-request', 'ping');
+    .catch(e => {
+        logger.error("Error processing versions for About page");
+    });
```
davemfish, line 155: Here's one I probably don't mind catching, because if the `About` page is missing some of it's data, it's no big deal. But I'd prefer logging the actual error message. I think it's in `e.message`.

src/static/report_a_problem.html
```diff
@@ -32,10 +32,17 @@
...
-      userDataPath = arg.userDataPath;
+
+    let userDataPath;
+
+    ipcRenderer.invoke('user-data')
+    .then(response => {
+      userDataPath = response;
+    })
+    .catch(e => {
+        logger.error("Error processing user data path.");
```
davemfish, line 43: Same deal as the about.html here.

davemfish (commented):
Hey @dcdenu4 , thanks for asking for my opinions of when to `catch`. I wasn't sure of the right thing to do, but then I found my opinions becoming more clear as I looked at these examples. Very helpful! Let me know what you think.

src/constants.js
```diff
@@ -3,10 +3,13 @@ import { ipcRenderer } from 'electron';
...
-ipcRenderer.on('variable-reply', (event, arg) => {
-  USER_DATA = arg.userDataPath;
+ipcRenderer.invoke('user-data')
+.then(response => {
+  USER_DATA = response;
 })
-ipcRenderer.send('variable-request', 'ping');
+.catch(e => {
+  console.error("Error processing user data path.");
+});
```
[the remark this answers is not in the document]
dcdenu4: I think I'm with you. I'd rather experience some of these `throws` in the wild and see what is happening before deciding on the best action for handling them and allowing the program to continue running. Done here: https://github.com/natcap/invest-workbench/pull/135/commits/43b8a43494ca0b0014002cb7af87855bb7e309eb

src/index.js
```diff
@@ -37,9 +40,12 @@ const render = async function render(investExe) {
...
+.then(response => {
   // render the App after receiving any critical data
   // from the main process
-  render(arg.investExe);
+  render(response.investExe);
 })
-ipcRenderer.send('variable-request', 'ping');
+.catch(e => {
+  logger.error("Error rendering the app.");
+});
```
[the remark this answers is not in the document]
dcdenu4: Would love to render the construction symbol! haha. Not sure that really makes sense for a desktop app!

Removed `catch` here: https://github.com/natcap/invest-workbench/pull/135/commits/43b8a43494ca0b0014002cb7af87855bb7e309eb

src/logger.js
```diff
@@ -7,10 +7,13 @@ let userDataPath = '';
...
-  ipcRenderer.on('variable-reply', (event, arg) => {
-    userDataPath = arg.userDataPath;
+  ipcRenderer.invoke('user-data')
+  .then(response => {
+    userDataPath = response;
   })
-  ipcRenderer.send('variable-request', 'ping');
+  .catch(e => {
+    console.error("Error processing user data path.");
+  });
```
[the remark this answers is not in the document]
dcdenu4: Throwing! https://github.com/natcap/invest-workbench/pull/135/commits/43b8a43494ca0b0014002cb7af87855bb7e309eb

src/static/about.html
dcdenu4: Good point, `e.message` it is: https://github.com/natcap/invest-workbench/pull/135/commits/6be3fcc4ec592f9cc8366d654a464439f8b1e245

src/static/report_a_problem.html
dcdenu4: Same: https://github.com/natcap/invest-workbench/pull/135/commits/6be3fcc4ec592f9cc8366d654a464439f8b1e245

dcdenu4:
Hmm, not sure why the test is failing now.
```
FAIL tests/binary_tests/puppet.test.js (12.395 s)
  × Run a real invest model (1798 ms)
  √ App re-launch will exit and focus on first instance (689 ms)

  ● Run a real invest model

    Evaluation failed: Error: Unable to find role="table"
```

If I run the test isolated locally, it passes fine. If I run `puppet.tests.js` locally in full, it hangs on the `App re-launch will exit` test, but the `Run a real invest model` test passes.

dcdenu4:
Alright @davemfish , I think this is ready for another look.

davemfish (approved):
Thanks, @dcdenu4 this looks good to me!

PR #139, opened 2021-04-09
cleanup the stream handlers in a WillUnmount. #138.

### LINKED ISSUE #138: intermittent memory-leak warnings from React
I've only seen this come up in tests, and only intermittently. But sometimes a test fails with a warning like this: 

```
2021-04-09T13:07:59.6049360Z     Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in %s.%s
2021-04-09T13:07:59.6051680Z 
2021-04-09T13:07:59.6052640Z       5 | console.error = function (message) {
2021-04-09T13:07:59.6053670Z       6 |   error.apply(console, arguments) // keep default behaviour
2021-04-09T13:07:59.6054730Z     > 7 |   throw (message instanceof Error ? message : new Error(message))
2021-04-09T13:07:59.6057170Z         |                                               ^
2021-04-09T13:07:59.6057860Z       8 | }
2021-04-09T13:07:59.6058820Z 
2021-04-09T13:07:59.6060280Z       at console.Object.<anonymous>.console.error (tests/jest.overrides.js:7:47)
2021-04-09T13:07:59.6062750Z       at printWarning (node_modules/react-dom/cjs/react-dom.development.js:67:30)
2021-04-09T13:07:59.6064490Z       at error (node_modules/react-dom/cjs/react-dom.development.js:43:5)
2021-04-09T13:07:59.6069050Z       at warnAboutUpdateOnUnmountedFiberInDEV (node_modules/react-dom/cjs/react-dom.development.js:23914:9)
2021-04-09T13:07:59.6071400Z       at scheduleUpdateOnFiber (node_modules/react-dom/cjs/react-dom.development.js:21840:5)
2021-04-09T13:07:59.6073350Z       at Object.enqueueSetState (node_modules/react-dom/cjs/react-dom.development.js:12467:5)
2021-04-09T13:07:59.6075050Z       at InvestTab.setState (node_modules/react/cjs/react.development.js:365:16)
2021-04-09T13:07:59.6077880Z       at Readable.<anonymous> (src/components/InvestTab/index.jsx:206:12)
```

In this particular case I think the `setState` is triggered after the unmount because we're not cleaning up the object that is holding a stream handler - which calls `setState` -- and there's a race condition where stderr is sent to that stream right before/after the test completes and component unmounts.

davemfish:
Fixes #138 

Here's a tiny PR that hopefully address why some of our tests fail intermittently (mainly on macos Actions, I think). It's a little hard to know for certain that this solved the problem, but these tests have run & re-run a few times now with this change, and they have passed.

This affects the `npm run tests` tests, but not the puppeteer tests that run in the Build step. They have their own unrelated reliability issues.

Here's an example of formerly failing tests:
https://github.com/davemfish/invest-workbench/runs/2306056762?check_suite_focus=true

### DIFF
--- a/src/components/InvestTab/index.jsx
+++ b/src/components/InvestTab/index.jsx
@@ -96,6 +96,11 @@ export default class InvestTab extends React.Component {
     }, () => { this.switchTabs('setup'); });
   }
 
+  componentWillUnmout() {
+    // this object holds stream handlers, so it's important to clean up
+    this.investRun = null;
+  }
+
   /** Write an invest args JSON file for passing to invest cli.
    *
    * Outsourcing this to natcap.invest.datastack via flask ensures


emlys (approved):
Looks good, thanks @davemfish! Just out of curiosity, is there any difference or reason to do the cleanup in `componentWillUnmount()` vs the `this.investRun.on('exit'...` section?

davemfish:
> Looks good, thanks @davemfish! Just out of curiosity, is there any difference or reason to do the cleanup in `componentWillUnmount()` vs the `this.investRun.on('exit'...` section?

@emlys I think I'm just following React guidance here. That particular lifecycle method seems to be designed for cleanup tasks like this. But I think you make a good point. The `WillUnmount` is kinda like the final safety-net - it's guaranteed to run that code.  We could remove the specific stream handlers as soon as we know we don't need them anymore, in the `.on('exit')` section. But we might still need the `WillUnmount` method because at least in tests there might be cases where the test completes before that .on('exit') is called, if it's alreay met all the `expects`.

emlys:
oh interesting, that `WillUnmount` is guaranteed to run seems like a great reason for the cleanup to go there. I'm gonna go ahead and merge this!

Issue #118, opened 2021-02-18
Workbench charset is undefined


phargogh:
As far as I can tell, the application doesn't defined a character encoding.  I'm pretty sure we're already guaranteeing that `invest serve` will always return UTF-8, but we should also be able to guarantee that the browser will interpret it as UTF-8 and not something else.

I'm pretty sure we can define the character encoding as UTF-8 and be safe.

emlys:
Fixed in #140

PR #140, opened 2021-04-12
Fix macOS instance lock crash

### LINKED ISSUE #137: App won't launch on macOS
As described in #133, both @phargogh and I experienced this. Running `npm start` fails for me as well so it's not just an issue with the built app.
Ran `git bisect` and it seems the problem originated in [this commit](https://github.com/natcap/invest-workbench/commit/5179b3093a6f3cfc390836a70a53435311412bd9). Simply requesting the lock in the line `const gotTheLock = app.requestSingleInstanceLock();` triggers it.

Relevant info:
https://github.com/electron/electron-osx-sign/issues/223#issuecomment-611129476
https://github.com/electron/electron/issues/15958
Not sure why other people are having this issue specifically with Mac App Store builds. [This solution](https://github.com/electron/electron/issues/15958#issuecomment-447685065) may work for our case.

emlys:
Fixes #137 
As the [electron docs](https://www.electronjs.org/docs/api/app#apprequestsingleinstancelock) say, you'll only end up with multiple instances of the app on macOS if you start one or more of them from the command line. The normal ways of opening an app automatically open the existing instance if one exists. I think it's fine to leave it this way, anyone who's opening it from the command line can probably understand the `Address already in use` error.

This has probably always been the case, but I just noticed if you `npm start` the app, then shut it down by pressing control+C, it doesn't shut down the flask server. Starting the app again will get the `Address already in use` error. Properly quitting the app does shut down the server.

Also made a few changes to address #118:
- Double checked that the server responds in UTF-8 and the responses are decoded as UTF-8
- Set the `charset` tag to `UTF-8` in `index.html`. This is probably the default, but it would be up to chrome to decide, best to be explicit.
- Set the `defaultEncoding` property when instantiating `BrowserWindow`s.

I'm not sure if these changes had any effect because I was able to `console.log` Russian characters and display them in the model list before making any changes. I wouldn't expect this to work because `defaultEncoding` [defaults](https://www.electronjs.org/docs/api/browser-window#new-browserwindowoptions) to `ISO-8859-1`.

Also made a super minor change to the wording on the "Report a problem" page because it could sound like the opposite meaning.

### DIFF
--- a/tests/binary_tests/puppet.test.js
+++ b/tests/binary_tests/puppet.test.js
@@ -73,7 +73,7 @@ beforeAll(async () => {
     console.log(`${data}`);
   });
   // so we don't make the next fetch too early
-  await new Promise((resolve) => setTimeout(resolve, 5000));
+  await new Promise((resolve) => setTimeout(resolve, 20000));
   const res = await fetch(`http://localhost:${PORT}/json/version`);
   const data = JSON.parse(await res.text());
   BROWSER = await puppeteer.connect({
--- a/tests/binary_tests/puppet.test.js
+++ b/tests/binary_tests/puppet.test.js
@@ -167,18 +167,25 @@ test('Run a real invest model', async () => {
 // The test is that the subprocess exits within a certain reasonable timeout.
 // Also verify that window 1 has focus.
 test('App re-launch will exit and focus on first instance', async () => {
-  await waitFor(() => {
-    expect(BROWSER.isConnected()).toBeTruthy();
-  });
+  if (process.platform === 'win32') {
+      await waitFor(() => {
+      expect(BROWSER.isConnected()).toBeTruthy();
+    });
 
-  // Open another instance of the Workbench application.
-  // This should return quickly.  The test timeout is there in case the new i
-  // process hangs for some reason.
-  const otherElectronProcess = spawnSync(
-    `"${BINARY_PATH}"`, [`--remote-debugging-port=${PORT}`],
-    { shell: true }
-  );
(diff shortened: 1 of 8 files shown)


emlys:
I merged in the changes from #116 which worked several times on that branch but now I'm getting the same problem here 😭 The macOS puppeteer test keeps hanging

davemfish:
> I merged in the changes from #116 which worked several times on that branch but now I'm getting the same problem here 😭 The macOS puppeteer test keeps hanging

I feel your pain. I've been struggling a lot with pptr lately in Windows GHA too.

tests/binary_tests/puppet.test.js
```diff
@@ -167,18 +167,25 @@ test('Run a real invest model', async () => {
...
+      `"${BINARY_PATH}"`, [`--remote-debugging-port=${PORT}`],
+      { shell: true }
+    );
+
+    // When another instance is already open, we expect an exit code of 1.
+    expect(otherElectronProcess.status).toBe(1);
+  } else {
+    // Single instance lock caused the app to crash on macOS, and also
+    // is less important because mac generally won't open multiple instances
+    console.log("Skipping this test because we're not on Windows");
```
emlys, line 188: I debated moving this test into a separate file and actions step that would only run on windows but did this because it's simpler. Do you think it would be better as a separate actions step?

davemfish: I'm good with this!

davemfish (approved):

PR #143, opened 2021-04-23
Feature/54 sample data distribution

### LINKED ISSUE #54: invest sample data distribution
Right now, invest sample data is distributed only to Windows users during invest installation. The distribution is wrapped up in the NSIS installer.

For the workbench, we probably could still use an NSIS installer to install the workbench and offer sample data downloads to users. But, I think it would be really nice to design a cross-platform way to distribute sample data. 

If we want the sample data included with downloads of the workbench, it might be possible to do that with electron-builder. Though the data might end up buried somewhere in the application's installation folder (.App on macOS). 

**So it might be nicer - and simpler - to design a UI feature for fetching sample data on demand.**

davemfish:
This PR fixes #54 

The app now uses a file token in the `'userData'` directory to determine if the app has ever been run before. If it has not, the download data modal is shown immediately on page load. If the app has been run before, this modal is not shown, but it can always be opened via the settings menu.

The modal displays the list of available invest sample data for the specific version of invest that is running. This list is kept in a new file - `src/sampledata_registry.json` - and is updated to stay current with the invest version listed in `package.json` each time we run `npm run fetch-invest`.

Downloads triggered by the user from the Modal are handled by the electron API in the main process. And some data is sent back to the renderer in order to display a progress bar. Zip files are extracted in place and then removed after downloaded. 

The user selects a directory to save all the selected zipfiles, and that directory path is cached in the `investSettings` db. So far we don't use that data, but we expect that we will use it for another feature that enables easy loading of a model's sample datastack. 

Missing from this PR:
Tests for the new main process functions & handlers. I'm putting this off based on our plan to refactor the whole main process side of the app with the goal of making it more modular and easier to test.


src/app.jsx
```diff
@@ -198,53 +215,85 @@ export default class App extends React.Component {
...
+              batchUpdateArgs={this.batchUpdateArgs}
             />
-          </TabPane>
-          {investTabPanes}
-        </TabContent>
-      </TabContainer>
+            {
+              // don't render until after we fetched the data
+              (investSettings)
+                ? (
```
dcdenu4, line 261: Is this to avoid a user clicking the download button and bringing up the download modal while it's currently working?

src/components/DataDownloadModal/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+      </Expire>
+    );
+  }
+  return (
+    <ProgressBar
+      animated
+      max={1}
+      now={nComplete / nTotal}
+      label={`Downloading ${nComplete + 1} of ${nTotal}`}
+    />
```
dcdenu4, line 221: I think the progress bar is fine and works well for downloading individual model data. However, when selecting all and downloading it just sits with 1/1 and the bar to the far left... Doesn't really provide a visual indication of progress. But, I don't really think it's a big deal.

src/styles/style.css
```diff
@@ -9,6 +9,7 @@ body {
 /*Top Navigation*/
 
 .navbar {
+	align-items: baseline;
```
dcdenu4, line 12: So this is a flex item option that aligns everything to the same horizontal plane. Is that right?

src/main_helpers.js
```diff
@@ -87,3 +89,58 @@ export function createPythonFlaskProcess(investExe) {
...
+        if (/\/$/.test(entry.fileName)) {
+          // if entry is a directory
+          fs.mkdir(path.join(extractToDir, entry.fileName), (err) => {
+            if (err) {
+              if (err.code === 'EEXIST') { } else logger.error(err);
+            }
+            zipfile.readEntry();
+          });
+        } else {
+          console.log(entry.fileName);
```
dcdenu4, line 109: Is this `console.log` left here intentionally? Just checking.

src/main_helpers.js
```diff
@@ -87,3 +89,58 @@ export function createPythonFlaskProcess(investExe) {
...
+            });
+            const writable = fs.createWriteStream(path.join(
+              extractToDir, entry.fileName
+            ));
+            readStream.pipe(writable);
+          });
+        }
+      });
+      zipfile.on('close', () => {
+        console.log('CLOSED')
```
dcdenu4, line 123: Also checking this `console.log` is intended.

src/main_helpers.js
```diff
@@ -87,3 +89,58 @@ export function createPythonFlaskProcess(investExe) {
...
+ * @returns {boolean}
+ */
+export function checkFirstRun() {
+  const userDataPath = app.getPath('userData');
+  const hasRunTokenPath = path.join(userDataPath, 'app-has-run-token');
+  try {
+    if (fs.existsSync(hasRunTokenPath)) {
+      return false;
+    }
+    fs.writeFileSync(hasRunTokenPath, '');
```
dcdenu4, line 141: So we're just using a blank file as the indicator of whether a user has run the workbench. This seems fine, but I'm curious if there's a reason we wouldn't use an established settings tables like invest settings? Maybe the way we load and use our other settings, this is easier and less obstructive?

src/main.js
```diff
@@ -157,12 +166,66 @@ const createWindow = async () => {
...
   });
+
+  // Setup handlers & defaults to manage downloads.
+  // Specifically, invest sample data downloads
+  let downloadDir;
+  let downloadLength;
+  const downloadQueue = [];
+  ipcMain.on('download-url', async (event, urlArray, directory) => {
+    logger.debug(`${urlArray}`);
+    downloadDir = directory;
```
dcdenu4, line 177: Is `downloadDir` used at all here?

src/main.js
```diff
@@ -157,12 +166,66 @@ const createWindow = async () => {
...
+    downloadQueue.push(...urlArray);
+    downloadLength = downloadQueue.length;
+    mainWindow.webContents.send(
+      'download-status',
+      [(downloadLength - downloadQueue.length), downloadLength]
+    );
+    urlArray.forEach((url) => mainWindow.webContents.downloadURL(url));
+  });
+
+  mainWindow.webContents.session.on('will-download', (event, item) => {
```
dcdenu4, line 187: Where does this get called from? A quick `grep` didn't turn it up, but maybe I'll find it in some of the files I've yet to look through.

src/components/SettingsModal/SettingsStorage.js
```diff
@@ -31,9 +33,34 @@ export function getDefaultSettings() {
  */
 export async function getSettingsValue(key) {
   const value = await investSettingsStore.getItem(key);
+  if (!value) {
+    return getDefaultSettings()[key];
```
dcdenu4, line 37: Hey Dave, I'm just curious how you found this necessary? Is this just to avoid a case where we've somehow corrupted the settings object?

src/components/SettingsModal/SettingsStorage.js
```diff
@@ -31,9 +33,34 @@ export function getDefaultSettings() {
...
+ * @returns {Object} - key: value pairs of settings
+ */
+export async function getAllSettings() {
+  try {
+    const promises = [];
+    const keys = Object.keys(getDefaultSettings());
+    keys.forEach((key) => {
+      promises.push(getSettingsValue(key));
+    });
+    const values = await Promise.all(promises);
```
dcdenu4, line 53: Nice, I like workflow of requesting all the values and then awaiting responses for all the promises.

src/components/SettingsModal/SettingsStorage.js
```diff
@@ -31,9 +33,34 @@ export function getDefaultSettings() {
...
+    const settings = Object.fromEntries(keys.map(
+      (_, i) => [keys[i], values[i]]
+    ));
```
dcdenu4, line 56: Cool.

src/components/DataDownloadModal/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+    const { url } = sampledataRegistry.Models[modelName];
+    if (event.target.checked) {
+      selectedLinksArray.push(url);
+      dataListCheckBoxes[modelName] = true;
+    } else {
+      selectedLinksArray = selectedLinksArray.filter((val) => val !== url);
+      dataListCheckBoxes[modelName] = false;
+    }
+    this.setState({
+      allDataCheck: false,
```
dcdenu4, line 103: Is this necessarily true? If I have all data selected and then toggle on and off a model, would it be convenient if `allDataCheck` was still `true`?

src/components/DataDownloadModal/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+  show: PropTypes.bool.isRequired,
+  closeModal: PropTypes.func.isRequired,
+  storeDownloadDir: PropTypes.func.isRequired,
+};
+
+export function DownloadProgressBar(props) {
+  const [nComplete, nTotal] = props.downloadedNofN;
+  if (nComplete === nTotal) {
+    return (
+      <Expire delay={props.expireAfter}>
```
dcdenu4, line 210: I'm a little confused about what the Expire component is doing. Is it easy to summarize?

scripts/fetch_invest_binaries.js
```diff
@@ -53,10 +56,129 @@ function download(src, dest) {
...
+        }
+      });
     });
   }).on('error', (e) => {
     console.log(e);
   });
 }
 
-download(urladdress, DESTFILE);
+/** Update a JSON registry of sampledata zipfiles available on GCS.
```
dcdenu4, line 79: This run down is great.

dcdenu4 (commented):
Hey @davemfish this looks great. I really like the registry addition and how it's worked into the `fetch-invest` and build scripts. I just had a few questions, some clarifying and some for my own curiosity. Thanks!

src/app.jsx
davemfish: No it's not strictly related to this PR. I think I hit the same problem you did a while back, the `SettingsModal` was rendered at least once before it's props (`investSettings`) were fetched. And so `SettingsModal` was being called with unexpected data. It meant having to wait to `find` elements in tests, which is fine, but I think this is better since there is no good that can come from rendering `SettingsModal` without `investSettings`.

src/components/DataDownloadModal/index.jsx
davemfish: Right, it does look a bit janky. We could also track progress by bytes downloaded instead of number of files. But the problem is we actually download and then extract, and at least for me, extraction takes longer. So I wanted the progress bar to wait for both steps. Anyway maybe we can make this look a little cleaner.

Sidenote: the download actually starts right away when the button is clicked, even before the user selects the save location. So the download is often already done before the save location is picked! Not certain how that works, but pretty cool.

src/styles/style.css
davemfish: Right. I added the progress bar to this area, so must have needed to fix alignment after that.

src/main_helpers.js
davemfish: Nope, thanks!

src/main_helpers.js
davemfish: I actually did exactly that - `investSettings` - for like 90% of the development of this feature, then changed it and things got much easier! The main thing is to manage this from `main` rather than `renderer`. This way `App` can simply be told via props whether it's the first run or not, it doesn't have to do any logic during rendering to figure that out. And it makes it way easier to setup front-end tests. No need to do setup with the mocked `localForage` in order to control this. Just render <App> with/without the `isFirstRun` prop.

src/main.js
davemfish: Yeah, it's not obvious, but it's assigned here and used in the 'will-download' listener.

src/main.js
davemfish: Good question, 'will-download' is triggered by `mainWindow.webContents.downloadURL(url)`. 

This seems a bit convoluted, but I think it's the correct use of the API. The 'will-download' event needs to be fired from the `main` process because its emitter is on `mainWindow.webContents`. But the initial user-event of course comes from the `renderer`, so we have that 'download-url' listener as kind of the go-between channel between the renderer and the 'will-download' callback. 

I think this is another good reason to refactor `main` into smaller modules, so it can be easier to see which things are closely related.

src/components/SettingsModal/SettingsStorage.js
davemfish: I'm not sure it strictly is necessary, though these settings are never meant to undefined, so if somehow they are, I think it's good that `App` never knows that and always gets valid values to render. 

Anyway this is actually the same logic that was already in place. It used to be inside the `app.jsx` componentDidMount. I pulled it out of there so we never have to think about the internals of `SettingsStorage` when we're dealing with `app.jsx`. 

Here's those changes: https://github.com/natcap/invest-workbench/pull/143/files/1833d7c1b42aabad24befd6d31da372145bdfbf2#diff-73ba1bb45657b12f529b48ba4515bc802efe0eb6184f56c364c6f5c9753b7441R57

src/components/SettingsModal/SettingsStorage.js
davemfish: I recall being a little surprised there was no localForage API for retrieving the whole db, so this seems to be the conventional way to do it, one key at a time.

src/components/DataDownloadModal/index.jsx
davemfish: Yeah good question. That would just require checking if all the `dataListCheckBoxes` are `true` on each click. I'm not sure it matters to the user, but there is actually different behavior. If "Select All" is checked, we download one big zipfile that includes everything, as opposed to downloading all the individual folders. I think I did that just because that one big zipfile exists as part of the invest build & deploy process. But that doesn't mean we have to use it.

src/components/DataDownloadModal/index.jsx
davemfish: Quick fix for this was `now={(nComplete + 1) / nTotal}`

Still not very useful as progress indicator for a single file, but at least it looks more professional.

davemfish:
> Hey @davemfish this looks great. I really like the registry addition and how it's worked into the `fetch-invest` and build scripts. I just had a few questions, some clarifying and some for my own curiosity. Thanks!

Thanks for reviewing @dcdenu4 ! I pushed a couple minor changes. I didn't touch the "Select All" behavior yet, but could. I'm kinda wondering if it makes sense to always download the batch of individual zipfiles instead of the one big one - even when Select All. Can you think of any reasons to go either way? We'd solve our "useless progress bar problem" if we got rid of the "one big zip" download.

davemfish:
I also forgot to include the global "Base Data" in the registry, so just pushed those changes.

src/app.jsx
dcdenu4: Ah, right I see. When I read `don't render until after we fetched the data` I was in the mind set of download data.

src/components/DataDownloadModal/index.jsx
dcdenu4: Yea, that looks a lot better.

>So the download is often already done before the save location is picked!

That is super cool.

src/main_helpers.js
dcdenu4: Sounds good to me! Thanks for walking through that decision.

src/main.js
dcdenu4: Ahhh, yea, I see it now. Thanks.

src/main.js
dcdenu4: Ahhh interesting. Thanks for that. I'm so used to these emitters and listeners basically have the same corresponding names. Yea, I think the refactor of `main` will tidy things up.

dcdenu4:
Hey @davemfish

> Can you think of any reasons to go either way?

If there isn't a download speed advantage I can't think of a reason we shouldn't just stick with downloading individual zip files. I like the idea of only having one method too. Should we go that route in this PR?

davemfish:
> If there isn't a download speed advantage I can't think of a reason we shouldn't just stick with downloading individual zip files. I like the idea of only having one method too. Should we go that route in this PR?

Done! This does make the code simpler so I'd say it's a clear win. Back to you @dcdenu4

tests/app.test.js
```diff
@@ -798,15 +770,15 @@ describe('Download Sample Data Modal', () => {
     );
 
     const allCheckBoxes = await findAllByRole('checkbox');
-    // toggle off one checkbox so not all are selected
-    fireEvent.click(allCheckBoxes[3]); // just avoid [0] - the SelectAll box
+    // // toggle off one checkbox so not all are selected
```
dcdenu4, line 773: This is very silly but, extra comment syntax?

tests/app.test.js
```diff
@@ -798,15 +770,15 @@ describe('Download Sample Data Modal', () => {
...
+    // fireEvent.click(allCheckBoxes[3]); // just avoid [0] - the SelectAll box
```
dcdenu4, line 774: Can we just delete this line now?

dcdenu4 (changes requested):
Hey @davemfish great! I just noticed two trivial things in `app.test`

davemfish:
> Hey @davemfish great! I just noticed two trivial things in `app.test`

Thanks, I removed the commented code.

dcdenu4 (approved):
Thanks @davemfish . Looks good.

PR #150, opened 2021-06-22
Task/146 using babel --watch to unify dev and production builds

### LINKED ISSUE #146: transpile all front-end code before runtime
This is mentioned as part of #60 , but I think it makes sense to do it first as it's own isolated PR.

Right now, we have babel generating node-compliant javascript, but for the front-end parts of the app, we then rely on electron's nodeIntegration to run that nodejs javascript in the browser.

In order to set nodeIntegration: false and contextIsolation: true, we'll need to generate browser-compliant javascript, and webpack is the tool for that.

I don't know all the details of implementing this yet, but we'll need to deal with transpiling the code for devMode & for production. For production, this essentially the `npm run build` step. And for devMode, we currently rely on `@babel/register` to transpile modules on the fly as they are required/imported.

We might be able to keep using babel for the main process code that runs in node, but we may also decide to make a full switch to webpack and drop babel. At least I _think_ that will be an option.

davemfish:
This fixes #146 - though that issue turned out to be slightly incorrect, I used it to generally improve how we use `babel`.

The most important changes here are in `package.json` scripts and the readme, which describes a slightly new development process. The changes allow us to run the same `build/` dir code during development & production. And that means we have less `isDevMode`-type logic in the source code, which is nice. So `index.html` and `index.js` get simpler as a result.

This PR also integrates react-devtools chrome extension into the main electron startup process, so the react tabs appear in devtools instead of needing the developer to start react-devtools separately.


scripts/clean-build-dir.js
```diff
@@ -7,19 +13,6 @@ const glob = require('glob');
...
-  // clean before build just to remove any files that may
-  // have been removed from src/ code but are still in build/
-  // from a previous build.
-  clean();
-  // TODO: formerly we were doing other stuff here,
-  // and we might again in the future with work on
-  // https://github.com/natcap/invest-workbench/issues/146
-  // So we should clean this up as part of that issue.
-}
-
```
davemfish, line 22: Whoops just noticed a glaring error here. I removed the function call completely.

package.json
```diff
@@ -2,16 +2,17 @@
...
   "scripts": {
-    "start": "electron -r @babel/register .",
+    "start": "electron .",
     "lint": "eslint --cache --color --ext .jsx,.js src",
     "test": "jest --runInBand --testPathIgnorePatterns /tests/binary_tests/",
     "test-electron-app": "jest --runInBand --testMatch **/tests/binary_tests/puppet.test.js",
     "postinstall": "electron-builder install-app-deps",
     "fetch-invest": "node ./scripts/fetch_invest_binaries.js && git diff src/sampledata_registry.json",
-    "build": "node ./build.js clean && babel src -d build --verbose --config-file ./babel.config.js --copy-files",
+    "dev": "node ./scripts/clean-build-dir.js && babel src -d build --verbose --config-file ./babel.config.js --copy-files --ignore **/__mocks__/* --no-copy-ignored --watch -s both",
```
dcdenu4, line 14: This looks pretty cool. Liking that `watch` flag.

readme.md
```diff
@@ -7,51 +7,27 @@ InVEST models, and to be extensible to future models or common auxilary
...
-Or build your own local invest binaries using invest's `make binaries` and then
-copy the entire `dist/invest/` dir to this repo's `build/invest/`.  
-
+* `npm install`
+* `npm run fetch-invest`
+	+ fetches prebuilt invest binaries (see package.json invest property)
+	+ Alternatively, build your own local invest binaries:
+		* use invest's `make binaries`, then
+		* `cp -r invest/dist/invest/ invest-workbench/build/invest/`
+* `npm run dev` (this process stays live, do it in a separate shell)
```
dcdenu4, line 16: Okay, so with `npm run dev` and `npm start` below, we'll have two shells hanging out. Is that right?

readme.md
```diff
@@ -7,51 +7,27 @@ InVEST models, and to be extensible to future models or common auxilary
...
-Before running `npm start`, in a second shell launch `react-devtools` (or `npx react-devtools` if you don't have a global install). Then run `npm start`.  
-The electron app should communicate with the
-standalone react-devtools window. There's an oustanding issue in electron
-preventing react-devtools from integrating with the electron browser's devtools window. https://github.com/electron/electron/issues/23662
-
-## To package this app for distribution  
-
-`npm run fetch-invest` - downloads invest binaries and updates the sampledata_registry.json.
+#### To run javascript outside the electron runtime,
+but with the same ECMAscript features and babel configurations:
```
dcdenu4, line 20: Is this the desired markdown syntax you wanted? Just checking.

scripts/clean-build-dir.js
```diff
@@ -1,25 +1,16 @@
-'use strict';
+/* We can't rely on babel to clean it's output directory every time
```
dcdenu4, line 1: Thanks for this comment.

src/index.html
```diff
@@ -1,34 +1,31 @@
...
+    <link
+      rel="stylesheet"
+      href="styles/style.css"
+    />
+  </head>
 
+  <body>
+    <div id="App"></div>
+  </body>
+  <script src="./index.js" type="module">
```
dcdenu4, line 29: So clean!

src/main/main.js
```diff
@@ -143,7 +143,18 @@ export function main(argv) {
     }
   }
 
-  app.on('ready', createWindow);
+  app.on('ready', async () => {
```
dcdenu4, line 146: I'm not entirely sure how this works, but I can at least tell it's a lot simpler design than what we had for setting up dev mode.

dcdenu4 (approved):
Hey @davemfish , This looks awesome. Certainly there are things I don't understand, but this looks very clean and nicely set up. I'm excited to learn more about it in the future.

dcdenu4:
Oh, @davemfish , I did have some comments, so I approved but have not merged. Just in case I don't make it back here before the weekend.

readme.md
davemfish: Yes, that's right

readme.md
davemfish: more/less, yeah. I think it looks okay.

src/main/main.js
davemfish: This is just for loading the react-devtools extension in the browser. It follows that packages documentation about when exactly do that, relative to actually creating the BrowserWindow.

davemfish:
Thanks, @dcdenu4 . I'll go ahead and merge it!

PR #152, opened 2021-06-23
Task/145 organizing renderer files

### LINKED ISSUE #145: re-organize src directory to better separate main & renderer code
I've already done this part way by adding a `src/main/` folder as part of #136 . And I was going to add a `src/renderer/` and move our front-end code in their, but I don't want to totally pollute the diff for the #136 PR, so I'm just creating this issue instead. And later we can do a PR that is strictly devoted to moving files around, and possibly some renaming too.

We might also want to mimic that pattern in `tests/` with `tests/main/` and `tests/renderer/`.

Oh, and the reasoning: in prep for #60 and `contextIsolation`, it's important to make it obvious which code is being imported and run by which process. It's obvious for some files (`.jsx`), but less so for others (e.g. `utils.js`). So dividing our `src` code is one way to make that clearer.

davemfish:
This PR fixes natcap/invest-workbench#145. It creates a `src/renderer/` and `tests/renderer/` directories to match the structure we recently put in place for `main` process code.

I also untracked all the `ResultsTab/` files instead of moving them. That was overdue since we have not been using these files quite some time. I commented on natcap/invest#874 about how to get them back if/when we want to.


emlys (approved):
Looks pretty straightforward. Thanks, @davemfish!

Issue #87, opened 2021-01-22
Some minor UI issues


emlys:
A catch-all ticket for some small UI things that would improve the appearance but don't affect usability. Feel free to add anything else you notice!

- [x] More clearly indicate if an input is optional. In the QT implementation the label text would contain "(optional)". Since we're now getting the label text from the args spec, it should be possible to append "(optional)" to the label if `argsSpec.arg.required === false`.
- [x] Disabled inputs still sometimes have a red or green outline, and any text typed into them before disabling (or selected dropdown option) isn't grayed out. This makes it less obvious that they're disabled.
- [x] input labels have inconsistent capitalization
- [x] Dropdowns should expand to fill the available width like text fields do, or at least expand to fit their widest option.
- [x] The style for separating groups of inputs isn't super effective because when the longer input labels wrap onto multiple lines, that introduces extra space that looks just like the between-group margin. Try making the margin bigger or add a faint horizontal bar or background color.
- [x] Some UI elements don't handle resizing of window well. Should the window just scroll vertically / horizontally when shrunk too much?
- [x] There's a gray border line above the top model in the list of models. IMO there should only be borders between models.
- [ ] It's inconsistent which text is selectable. If you "Select All" on the home tab, the model names aren't selected. If you "Select All" on a model tab, "User's Guide" and "Frequently Asked Questions" are selected but "Load parameters from file", "Save to JSON", "Save to Python script" are not.
- [x] Default hover text appears on all elements and is not helpful. We should remove it and maybe add it back for elements that would really benefit from it.
- [x] Open Workspace button opens a native OS dialog, but on Windows it does not display on top of other open windows. So it's easy to think that the button didn't work.
- [ ] ~~input text fields should display the right-most end of their text content when they contain text that is too long to display completely. That way the more important part of the filepaths is always visible. This was implemented at some point but seems to no longer be working.~~ Moving this to a new issue.

davemfish:
Great list!

> If you hold down a key in a text input field, it only types the character once. Not really a problem, but different from the expected behavior in a form.

This one I'm not seeing in `main`, if I'm interpreting it correctly.

emlys:
@davemfish it turns out that is just a default macOS thing that I'd never noticed before 😬

davemfish:
> @davemfish it turns out that is just a default macOS thing that I'd never noticed before 😬

Woah, that's a new one!

davemfish:
[...]

PR #154, opened 2021-07-01
Feature/122 cleanup tabs styles

### LINKED ISSUE #122: Improve readability & intuitiveness of the Sidebar link titles
* FAQ - might be better as Frequently Asked Questions
* Log - the disabled style is pretty subtle and it might not be obvious what that button is for until after the model starts running. We could gray it out much more, and maybe expand the name to "View Log" or "View Model Log"
* User's Guide & FAQ - both could benefit from an icon that indicates this link opens in a new browser ( we could leave this out until after #102 to avoid changing it multiple times)

If it's useful to add further explanation to any of these items, we could do so in some hover text.

davemfish:
Various style & layout improvements here, mainly focused on keeping things clean & tidy at varying window widths. Fewer items are allowed to wrap, the overall window has a strict minimum 800 pixel width. Most of the attention went to items in the top navbar, and then the two main columns on each of the tabs.

This Fixes #122 and addresses a couple items in #87

### DIFF
--- a/tests/renderer/app.test.js
+++ b/tests/renderer/app.test.js
@@ -205,7 +205,9 @@ describe('Various ways to open and close InVEST models', () => {
     expect(tab3EventKey).not.toEqual(tab1EventKey);
 
     // Click the close button on the middle tab
-    const tab2CloseButton = await within(tab2).getByRole('button', { name: /x/ });
+    const tab2CloseButton = await within(tab2.closest('.nav-item'))
+      .getByRole('button', { name: /x/ });
+    // const tab2CloseButton = await within(tab2).getByRole('button', { name: /x/ });
     fireEvent.click(tab2CloseButton);
     // Now there should only be 2 model tabs open
     modelTabs = await findAllByRole('tab', { name: /Carbon/ });
@@ -215,7 +217,8 @@ describe('Various ways to open and close InVEST models', () => {
     expect(tab1.classList.contains('active')).toBeFalsy();
 
     // Click the close button on the right tab
-    const tab3CloseButton = await within(tab3).getByRole('button', { name: /x/ });
+    const tab3CloseButton = await within(tab3.closest('.nav-item'))
+      .getByRole('button', { name: /x/ });
     fireEvent.click(tab3CloseButton);
     // Now there should only be 1 model tab open
     modelTabs = await findAllByRole('tab', { name: /Carbon/ });
@@ -225,7 +228,8 @@ describe('Various ways to open and close InVEST models', () => {
     expect(homeTab.classList.contains('active')).toBeFalsy();
 
     // Click the close button on the last tab
-    const tab1CloseButton = await within(tab1).getByRole('button', { name: /x/ });
+    const tab1CloseButton = await within(tab1.closest('.nav-item'))
+      .getByRole('button', { name: /x/ });
     fireEvent.click(tab1CloseButton);
     // Now there should be no model tabs open.
     modelTabs = await queryAllByRole('tab', { name: /Carbon/ });
@@ -540,6 +544,7 @@ describe('InVEST subprocess testing', () => {
       findByText,
       findByLabelText,
       findByRole,
+      getByRole,
       queryByText,
       unmount,
     } = render(<App />);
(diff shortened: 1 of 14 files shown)


davemfish:
Here's a before and after screenshots to show some new styling of the tabs. (These are from puppeteer in GHA and that's why they don't display the window frame properly)

Before:
invest-workbench-4-complete-setup-form

After:
invest-workbench-4-complete-setup-form

And another after showing how multiple tabs now shrink as needed instead of having a horizontal scroll bar:
Capture

davemfish:
> After:

Just noticed the feedback font size got bigger here, that was inadvertent and I just pushed a change to make it smaller again.

tests/renderer/app.test.js
```diff
@@ -205,7 +205,9 @@ describe('Various ways to open and close InVEST models', () => {
     expect(tab3EventKey).not.toEqual(tab1EventKey);
 
     // Click the close button on the middle tab
-    const tab2CloseButton = await within(tab2).getByRole('button', { name: /x/ });
+    const tab2CloseButton = await within(tab2.closest('.nav-item'))
+      .getByRole('button', { name: /x/ });
+    // const tab2CloseButton = await within(tab2).getByRole('button', { name: /x/ });
```
dcdenu4, line 210: Good to remove this commented line?

dcdenu4 (changes requested):
Hey @davemfish this looks really good. I tested out your branch locally and everything resized very nicely. The only thing I noticed was that the "LOG" wording hasn't been changed. I think in one of the linked issues, it mentioned this being a little confusing and that maybe "View Log" or something might make more sense. Thoughts?

dcdenu4:
>Just noticed the feedback font size got bigger here, that was inadvertent and I just pushed a change to make it smaller again.

Great! I did notice the font was bigger when running a model, but I didn't mind it haha. Hopefully that doesn't mean I'm getting old.

tests/renderer/app.test.js
davemfish: Thank you!

davemfish:
> > Just noticed the feedback font size got bigger here, that was inadvertent and I just pushed a change to make it smaller again.
> 
> Great! I did notice the font was bigger when running a model, but I didn't mind it haha. Hopefully that doesn't mean I'm getting old.

Yeah this is something I still don't fully understand. We're using `rem` units for fonts everywhere, and I think the baseline font size must have changed for the text in the setup form, but I don't know exactly why, maybe as a side-effect of some of the other changes there. Anyway I agree it's suitable as-is and can easily be adjusted more.

davemfish:
> Hey @davemfish this looks really good. I tested out your branch locally and everything resized very nicely. The only thing I noticed was that the "LOG" wording hasn't been changed. I think in one of the linked issues, it mentioned this being a little confusing and that maybe "View Log" or something might make more sense. Thoughts?

Thanks for trying it out! I thought "Log" was a pretty good label after all. It's one of those things where the very first time you see it you might not know what it's about. But it's also disabled at that point and it enables and switches to it automatically when the user clicks "Run". And from that point on, I think the user will know what "Log" means.

davemfish:
Hmm we're getting annoying npm errors on the `npm install` step our workflows, before it even tries installing anything. It's been an intermittent problem. I just bumped the node version in the workflow files in case that helps.

dcdenu4:
hmmm now I'm seeing a mac failure from this test:

https://github.com/natcap/invest-workbench/runs/3001910418?check_suite_focus=true

davemfish:
> hmmm now I'm seeing a mac failure from this test:
> 
> https://github.com/natcap/invest-workbench/runs/3001910418?check_suite_focus=true

Yeah I just made #155 for that. I don't think it's related to these changes.

dcdenu4 (approved):
Approved.

dcdenu4:
>Hmm we're getting annoying npm errors on the npm install step our workflows, before it even tries installing anything. It's been an intermittent problem. I just bumped the node version in the workflow files in case that helps.

Judging by the tests this either helped or didn't show itself. So I'll go ahead and merge as it's not related to this PR.

Issue #112, opened 2021-02-16
Clicking "Open" on an untouched model inputs pane opens a new inputs pane


phargogh:
To reproduce:

* Open the workbench
* Open a model (I used Fisheries)
* Click the "open" button to attempt to load a set of input parameters into the existing pane
* Select the parameter set
* Observe that a new pane was opened, leaving the original, empty, pane open.

The behavior I'd expect is that if:

* The most recently-opened model has not been altered and
* The datastack selected is for the same model as what was just opened

Then it might be nice to load the parameters into the existing pane.  It's not a problem to manually close the existing tab down, but it is a slightly unexpected behavior.

davemfish:
Yeah, I hear you. Over on #98 we talked about maybe having the "Open" button only accessible from the home screen to avoid this. Or like you suggest, we could probably figure out logically if it should open a new tab or populate the already-open one. 

Let's address this issue and #98 at the same time.

davemfish:
This was fixed in #157

Issue #113, opened 2021-02-16
Graphically display model execution status in model tabs?


phargogh:
Since it's possible to run multiple InVEST models in parallel and in separate tabs, what if we were to display some sort of visual indicator of model run status (model executing/failed/completed) in the tab?

While running a few models in parallel, it was a little inconvenient to have to keep switching between tabs to see what the state of each model's execution was.

davemfish:
Great idea! That status is also viewable from the home tab on each job's card. I agree we should display it visually on the tab, probably not with more text, but maybe an ✔️ ❌ :spinner: And the same cue can be applied to those recent job cards.

dcdenu4:
✔️ 🟡  🚫

PR #157, opened 2021-07-13
Feature/98 load parameters button vs open button

### LINKED ISSUE #98: Add load option next to save options
When a model tab is open there are save options in the left column. I would expect a load option there as well, that would file browse to a parameter file and update the current model inputs. The Load button up top creates a new `InvestJob` and calls `openInvestModel(job)`.

Proposal
----------
Add `Load saved parameters` under the save options. Clicking would open a file browser and expect a `json` parameter file. On submit, the inputs would be batch updated like what occurs when a parameter json file is dropped on the model.

davemfish:
This PR fixes #98 and #112 . It moves the "Open" button out of the Navbar and onto the Home tab. And it creates a new "Load parameters from file" button available under the Setup menu next the "Save as..." buttons. The new load button has essentially the same behavior as a drag-and-drop onto the ArgsForm.

This PR also does a couple other misc items that became apparent:
* Fixed a bug introduced in #154 where the sample data download progress bar was not being displayed inline in the navbar during downloads.
* Moved the `onDragOver` property -- that we use to set the default dragover icon to "None" -- up to the top-level node that accepts it. That way we don't need to add this property to so many individual components. Our dragOver behavior still works everywhere it should. Presumably the properties on children override the one's on their parent.
* Replaced IPC channel strings with the constants that we defined during #147 - all the ipcRenderer calls still had hardcoded strings.

### DIFF
--- a/src/renderer/components/HomeTab/index.jsx
+++ b/src/renderer/components/HomeTab/index.jsx
@@ -8,7 +8,7 @@ import Container from 'react-bootstrap/Container';
 import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 
-import LoadButton from '../LoadButton';
+import OpenButton from '../OpenButton';
 import InvestJob from '../../InvestJob';
 
 const logger = window.Workbench.getLogger(__filename.split('/').slice(-2).join('/'));
@@ -150,7 +150,7 @@ class RecentInvestJobs extends React.PureComponent {
           <h4 className="d-inline-block">
             Recent runs:
           </h4>
-          <LoadButton
+          <OpenButton
             className="float-right"
             openInvestModel={this.props.openInvestModel}
             batchUpdateArgs={this.props.batchUpdateArgs}
--- a/src/renderer/components/HomeTab/index.jsx
+++ b/src/renderer/components/HomeTab/index.jsx
@@ -147,27 +147,27 @@ class RecentInvestJobs extends React.PureComponent {
     return (
       <Container>
         <div className="mb-1">
-          <h4 className="d-inline-block">
-            Recent runs:
-          </h4>
+          {recentButtons.length
+            ? (
+              <h4 className="d-inline-block">
+                Recent runs:
+              </h4>
+            )
+            : (
+              <div className="d-inline-block">
+                Try the <b>Open</b> button to setup a model from a sample 
+                datastack file (.json) or from an InVEST model's logfile (.txt)
+              </div>
+            )}
           <OpenButton
             className="float-right"
             openInvestModel={this.props.openInvestModel}
             batchUpdateArgs={this.props.batchUpdateArgs}
           />
         </div>
-        {recentButtons.length
-          ? (
-            <React.Fragment>
-              {recentButtons}
-            </React.Fragment>
-          )
-          : (
-            <div>
-              Try the <b>Open</b> button to setup a model from a sample 
-              datastack file (.json) or from an invest model's logfile (.txt)
-            </div>
-          )}
+        <React.Fragment>
+          {recentButtons}
+        </React.Fragment>
       </Container>
     );
   }
--- a/src/renderer/components/HomeTab/index.jsx
+++ b/src/renderer/components/HomeTab/index.jsx
@@ -8,6 +8,7 @@ import Container from 'react-bootstrap/Container';
 import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 
+import LoadButton from '../LoadButton';
 import InvestJob from '../../InvestJob';
 
 const logger = window.Workbench.getLogger(__filename.split('/').slice(-2).join('/'));
@@ -145,9 +146,16 @@ class RecentInvestJobs extends React.PureComponent {
 
     return (
       <Container>
-        <h4>
-          Recent runs:
-        </h4>
+        <div className="mb-1">
+          <h4 className="d-inline-block">
+            Recent runs:
+          </h4>
+          <LoadButton
+            className="float-right"
+            openInvestModel={this.props.openInvestModel}
+            batchUpdateArgs={this.props.batchUpdateArgs}
+          />
+        </div>
         {recentButtons.length
           ? (
             <React.Fragment>
@@ -156,8 +164,6 @@ class RecentInvestJobs extends React.PureComponent {
           )
           : (
             <div>
-              No recent InVEST runs yet.
-              <br />
               Try the <b>Open</b> button to setup a model from a sample 
               datastack file (.json) or from an invest model's logfile (.txt)
             </div>
--- a/src/renderer/components/OpenButton/index.jsx
+++ b/src/renderer/components/OpenButton/index.jsx
@@ -13,7 +13,7 @@ import { fetchDatastackFromFile } from '../../server_requests';
  * Render a button that loads args from a datastack, parameterset, or logfile.
  * Opens a native OS filesystem dialog to browse to a file.
  */
-export default class LoadButton extends React.Component {
+export default class OpenButton extends React.Component {
   constructor(props) {
     super(props);
     this.browseFile = this.browseFile.bind(this);
@@ -52,6 +52,6 @@ export default class LoadButton extends React.Component {
   }
 }
 
-LoadButton.propTypes = {
+OpenButton.propTypes = {
   openInvestModel: PropTypes.func.isRequired,
 };
--- a/src/renderer/components/OpenButton/index.jsx
+++ b/src/renderer/components/OpenButton/index.jsx
@@ -34,7 +34,7 @@ export default class OpenButton extends React.Component {
   }
 
   render() {
-    const tipText = 'Browse to a datastack (.json) or invest logfile (.txt)';
+    const tipText = 'Browse to a datastack (.json) or InVEST logfile (.txt)';
     return (
       <OverlayTrigger
         placement="left"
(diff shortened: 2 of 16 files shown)


src/renderer/components/SetupTab/index.jsx
```diff
@@ -206,6 +214,24 @@ export default class SetupTab extends React.Component {
...
   }
 
+  async loadParametersFromFile(filepath) {
+    const datastack = await fetchDatastackFromFile(filepath);
+
+    if (datastack.module_name === this.props.pyModuleName) {
+      this.batchUpdateArgs(datastack.args);
+    } else {
+      alert(
+        `Parameter/Log file for ${datastack.module_name} does not match this model: ${this.props.pyModuleName}`);
```
emlys, line 224: This alert message could be more user-friendly. For example, I got this alert:  _"Parameter/Log file for natcap.invest.crop_production_percentile does not match this model: natcap.invest.coastal_vulnerability"_ 
- Could we use the human-friendly model names here?
- I think we usually call it a "datastack file" and not a "parameter file". 
- I would take out the "`${this.props.pyModuleName}`" part. It may be useful for debugging but doesn't really help the user.
- If it's not too much work, it would be nice to say either "Datastack file" or "Log file" as appropriate for the type of file that was dropped.

emlys: This is minor, but the alert window shows up at the top of the app window. It would be nice to center it over the app.

src/renderer/components/SetupTab/index.jsx
```diff
@@ -354,12 +379,27 @@ export default class SetupTab extends React.Component {
...
+              loadParametersFromFile={this.loadParametersFromFile}
             />
           </Row>
           <Portal elId={sidebarSetupElementId}>
+            <OverlayTrigger
+              placement="right"
+              delay={{ show: 250, hide: 400 }}
+              overlay={(
+                <Tooltip>
+                  Browse to a datastack (.json) or invest logfile (.txt)
```
emlys, line 392: should this be "InVEST" for consistency?

src/renderer/components/SetupTab/index.jsx
```diff
@@ -354,12 +379,27 @@ export default class SetupTab extends React.Component {
...
-              pyModuleName={pyModuleName}
               updateArgValues={this.updateArgValues}
-              batchUpdateArgs={this.batchUpdateArgs}
+              loadParametersFromFile={this.loadParametersFromFile}
             />
           </Row>
           <Portal elId={sidebarSetupElementId}>
+            <OverlayTrigger
+              placement="right"
+              delay={{ show: 250, hide: 400 }}
```
emlys, line 389: Neat! The delay is subtle but really makes it

src/renderer/components/SetupTab/index.jsx
```diff
@@ -354,12 +379,27 @@ export default class SetupTab extends React.Component {
...
+                <Tooltip>
+                  Browse to a datastack (.json) or invest logfile (.txt)
+                </Tooltip>
+              )}
+            >
+              <Button
+                onClick={this.browseForDatastack}
+                variant="link"
+              >
+                Load parameters from file
```
emlys, line 400: When I made the app window narrower, there's a point where this text wraps to a second line and becomes centered, making it appear indented from the other sidebar options. It's not very long so maybe we could force the sidebar to always stay wide enough for this text?
image

tests/renderer/investtab.test.js
```diff
@@ -179,6 +217,28 @@ describe('Save InVEST Model Setup Buttons', () => {
...
     spy.mockRestore(); // restores to unmocked implementation
   });
+
+  test('Load Parameters Button: does nothing when canceled', async () => {
+    // this resembles the callback data if the dialog is canceled instead of 
+    // a save file selected.
+    const mockDialogData = {
+      filePaths: ['']
+    };
+    ipcRenderer.invoke.mockResolvedValue(mockDialogData);
```
emlys, line 227: Neat! I'm guessing we need this because it's not possible to test things in the choose-a-file window?

emlys (changes requested):
Thanks @davemfish, this is an improvement! It's nice we can set the `onDragOver` property in just one place.
I'm commenting on some drag-and-drop things that might be out of scope for this PR. It was mentioned in #98, but I can make a separate issue for it stuff if you prefer.

- I expected that I could drag-and-drop a datastack file onto the home screen to open a model, but this did not work
- Dragging and dropping a datastack file onto a model tab works great, and the visual cues make it clear when you're dropping it onto the whole tab (to fill all the fields) vs. dropping it onto a single filed.
- The alert message when the datastack doesn't match the model tab could be clearer

src/renderer/components/OpenButton/index.jsx
```diff
@@ -42,10 +42,9 @@ export default class LoadButton extends React.Component {
         overlay={<Tooltip>{tipText}</Tooltip>}
       >
         <Button
-          className="mx-3"
+          className={this.props.className}
```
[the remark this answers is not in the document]
davemfish: Good question, I went back and forth and I actually decided to assign the class on `OpenButton` because that's where the rest of the context for the DOM is. For example I thought "float-right" would be out of context here on `Button`, but would make more sense when also looking at the "d-inline-block" class assigned to the sibling node. But yeah, there's a tradeoff with having to pass it through props.

src/renderer/components/HomeTab/index.jsx
```diff
@@ -145,9 +146,16 @@ class RecentInvestJobs extends React.PureComponent {
...
     return (
       <Container>
-        <h4>
-          Recent runs:
-        </h4>
+        <div className="mb-1">
+          <h4 className="d-inline-block">
+            Recent runs:
+          </h4>
+          <OpenButton
```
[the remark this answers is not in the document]
davemfish: > To my eye this would look better with the Open button and corresponding "Try the Open button..." text placed above "Recent runs:". 

Yeah I see your point here. I'm changing this to replace the "Recent runs" text with the "Try the Open..." text when there are no recent runs. 

> I think it could make sense to have it even if there are recent runs to show. Maybe eventually, the "Try the Open button..." text could be part of a getting started/"tour" feature that only shows up the first time you use the app.

I don't really think the "Try the open button" button should always be present because once you've used it once, that text just becomes extra clutter.

src/renderer/components/SetupTab/index.jsx
davemfish: Thanks for the suggestions, here's the new text:

`Datastack/Logfile for Coastal Vulnerability model does not match this model.`

> This is minor, but the alert window shows up at the top of the app window. It would be nice to center it over the app.

Hmm it is centered on Windows. I guess the positioning is controlled by the OS? We may decide to switch away from `window.alert` and design our own at some point.

src/renderer/components/SetupTab/index.jsx
davemfish: Thanks! I believe I fixed this with `white-space: nowrap` on the button text, and `min-width: fit-content` on the sidebar column.

tests/renderer/investtab.test.js
davemfish: That's right!

src/renderer/components/SetupTab/index.jsx
emlys: Thanks, it works smoothly now

src/renderer/components/OpenButton/index.jsx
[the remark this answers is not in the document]
emlys: Okay, makes sense

src/renderer/components/SetupTab/index.jsx
```diff
@@ -206,6 +214,25 @@ export default class SetupTab extends React.Component {
...
   }
 
+  async loadParametersFromFile(filepath) {
+    const datastack = await fetchDatastackFromFile(filepath);
+
+    if (datastack.module_name === this.props.pyModuleName) {
+      this.batchUpdateArgs(datastack.args);
+    } else {
+      alert(
+        `Datastack/Logfile for ${datastack.model_human_name} model does not match this model.`
```
emlys, line 224: We have some inconsistency in the model human names where a couple of them already end in "Model", so it ends up repeating: _Datastack/Logfile for Crop Production: Percentile Model model does not match this model._ Maybe we can just leave out the second "model" there?

src/renderer/components/HomeTab/index.jsx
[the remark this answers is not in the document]
emlys: Thanks, it does look neater leaving out the "Recent runs:" text when there aren't any recent runs.

When there are recent runs, the placement of the Open button directly across from "Recent runs:" makes me think it would open a recent run, not open a datastack. But the tooltip text does make that clear. Maybe that's something to address in a future, general UX issue.

emlys (changes requested):
Thanks Dave! Just a couple more comments

src/renderer/components/HomeTab/index.jsx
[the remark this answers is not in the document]
davemfish: In a way it could be thought of as opening a recent run, as it can load from a logfile. Maybe it ought to have a 📂 to indicate it's a "browse" button rather than being associated with the available recent runs. Maybe I'll do that in #113 where I'm adding other emoji.

src/renderer/components/SetupTab/index.jsx
davemfish: Good catch. Ugh, inconsistent names. I removed the `model`.

davemfish:
> Thanks Dave! Just a couple more comments

Thanks, back to you!

emlys:
Looks good!

emlys (approved):

PR #163, opened 2021-08-02
Better management of invest job-related state

### LINKED ISSUE #113: Graphically display model execution status in model tabs?

### LINKED ISSUE #131: Model error alert box shows an unhelpful message
If a model ends in an error, the Open Workspace button & alert box are supposed to show the final error message from the model's log. Instead, the alert seems to always show "[21292] Failed to execute script cli".

I'm guessing that's some stderr emanating from the invest CLI. It's not something we should show to users.

davemfish:
This PR fixes #113 and fixes #131 . To accomplish #113, it was necessary for an invest job's status to be available to the `App` component, where the tabs are named. So it makes sense for all open invest job state to be managed in this component. Previously there was a bug where a `job` stored in `App.state` was being mutated by a child component, and it was only coincidence that the parent's re-render displayed the correct updated view. 

Now, all open (in-memory) invest job state is held in `App.jsx` and updated by child components through an `updateJobProperty` prop. Job metadata are saved to disk only when the invest process exits. And the `RecentInvestJobs` component is a view of that database - same as it was before but with minor enhancements to display the job's exit status/error message.

The `InvestJob` class was refactored slightly to eliminate the need for referencing a `metadata` property. So job properties are like `job.logfile` instead of `job.metadata.logfile`. The reason for the `metadata` property was to clearly separate that data which needed to be saved to the database, from other `job` properties which were actually instance methods. So to go along with this change, there no longer any instance methods, only static methods of `InvestJob`. I almost went so far as to get rid of the `InvestJob` class & constructor altogether - just using plain functions instead - but so far did I did not do that. Right now, the only purpose of the constructor is to enforce that a `new InvestJob` object always has a couple required properties.

Other minor changes:
* since there is now a spinner on the tab name while a model is running, I removed the spinner from the "Log" tab in the sidebar. 
* The `Alert` that appears in the sidebar while a model is running and after it completes is now in it's own module at `src/renderer/components/InvestTab/ModelStatusAlert/index.jsx`


davemfish:
I requested Doug on this one as I'm just cycling through the three of you, but I also vaguely described some of the problems I was working on here to @phargogh recently, in case you're interested in looking as well!

src/main/setupInvestHandlers.js
```diff
@@ -143,11 +143,16 @@ export function setupInvestRunHandlers(investExe) {
     // in which case it's useful to logger.debug too.
     investRun.stderr.on('data', (data) => {
       logger.debug(`${data}`);
-      event.reply(`invest-stderr-${channel}`, `${data}${os.EOL}`);
+      // The PyInstaller exe will always emit a final 'Failed ...' message
+      // after an uncaught exception. It is not helpful to display to users
+      // so we filter it out here.
+      const dat = `${data}`
+        .split(os.EOL)
+        .filter((line) => !line.match(/\[[0-9]+\] Failed to execute script/))
```
dcdenu4, line 151: So this just removes the PyInstaller `Failed` message from any stderr we want to report back?

src/renderer/InvestJob.js
```diff
@@ -43,83 +43,82 @@ export default class InvestJob {
...
+        job.argsValues.resultsSuffix
+      );
     }
     const isoDate = new Date().toISOString().split('T')[0];
     const localTime = new Date().toTimeString().split(' ')[0];
-    this.metadata.humanTime = `${isoDate}  ${localTime}`;
-    let sortedKeys = await investJobStore.getItem(HASH_ARRAY_KEY);
-    if (!sortedKeys) {
+    job.humanTime = `${isoDate} ${localTime}`;
+    let sortedWorkspaceHashes = await investJobStore.getItem(HASH_ARRAY_KEY);
```
dcdenu4, line 71: I noticed you switched from `sortedKeys` to `sortedWorkspaceHashes`. Would it be consistent to also use this naming above in the static functions? It's not necessary, but I thought maybe it'd be nice if we were taking the time to use a more thoughtful name.

src/renderer/InvestJob.js
```diff
@@ -43,83 +43,82 @@ export default class InvestJob {
...
-  }
-
-  setWorkspaceHash() {
-    if (this.metadata.argsValues.workspace_dir
-        && this.metadata.modelRunName) {
-      this.metadata.workspaceHash = crypto.createHash('sha1').update(
-        `${this.metadata.modelRunName}
-         ${JSON.stringify(this.metadata.argsValues.workspace_dir)}
-         ${JSON.stringify(this.metadata.argsValues.results_suffix)}`
+  static getWorkspaceHash(modelRunName, workspaceDir, resultsSuffix) {
```
dcdenu4, line 46: So `getWorkspaceHash` is just a name bump from `setWorkspaceHash`. Just curious why you felt it was more a 'getter' instead of a 'setter'?

src/renderer/app.jsx
```diff
@@ -11,6 +11,7 @@ import Nav from 'react-bootstrap/Nav';
 import Button from 'react-bootstrap/Button';
 import Row from 'react-bootstrap/Row';
 import Col from 'react-bootstrap/Col';
+import Spinner from 'react-bootstrap/Spinner';
```
dcdenu4, line 14: I tested out the version locally and the spinner worked great.

src/renderer/app.jsx
```diff
@@ -177,28 +199,53 @@ export default class App extends React.Component {
...
 
     const investNavItems = [];
     const investTabPanes = [];
-    openJobs.forEach((job) => {
+    openNavIDs.forEach((id) => {
+      const job = openJobs[id];
+      let statusSymbol;
+      switch (job.status) {
+        case 'success':
+          statusSymbol = '\u{2705}'; // green check
```
dcdenu4, line 215: Oh this is pretty cool. My first thought would have been to dig out a symbol / icon library for this.

src/renderer/components/InvestTab/index.jsx
```diff
@@ -110,58 +110,66 @@ export default class InvestTab extends React.Component {
...
       // Windows taskkill yields exit code 1
       // Non-windows process.kill yields exit code null
       const status = (code === 0) ? 'success' : 'error';
-      this.setState({
-        jobStatus: status
+      let finalTraceback = '';
+      if (this.state.logStdErr) {
+        let i = 1;
+        while (!finalTraceback) {
+          [finalTraceback] = `${this.state.logStdErr}`
```
dcdenu4, line 141: Just trying to walk through this while loop. If we have some stderr we'll parse it by line until `finalTraceback` will eventually be empty? what exactly is the `-1 * i` splice doing?

src/main/setupInvestHandlers.js
davemfish: That's it!

src/renderer/InvestJob.js
davemfish: Good idea!

src/renderer/InvestJob.js
davemfish: Now it returns the value, whereas before it actually set the value.

src/renderer/app.jsx
davemfish: I think I went with your suggested emoji in #113 !

src/renderer/components/HomeTab/index.jsx
```diff
@@ -134,9 +134,12 @@ class RecentInvestJobs extends React.PureComponent {
             </Card.Title>
             <Card.Text>{job.description || <em>no description</em>}</Card.Text>
             <Card.Footer className="text-muted">
-              <span>{job.humanTime}</span>
-              <span className="float-right">
-                <em>{job.status !== 'success' ? job.status : ''}</em>
+              <span className="timestamp">{job.humanTime}</span>
+              <span className="status-traceback">
+                {(job.status === 'success'
+                  ? '\u{2705}'
```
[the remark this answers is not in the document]
davemfish: That's right

src/renderer/components/InvestTab/index.jsx
```diff
@@ -49,18 +53,14 @@ async function investGetSpec(modelName) {
 export default class InvestTab extends React.Component {
   constructor(props) {
     super(props);
-
     this.state = {
       activeTab: 'setup',
       modelSpec: null, // ARGS_SPEC dict with all keys except ARGS_SPEC.args
       argsSpec: null, // ARGS_SPEC.args, the immutable args stuff
+      uiSpec: null,
```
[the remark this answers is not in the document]
davemfish: It was missing from the constructor here but we are actually setting a value for it during `componentDidMount`. I guess we don't strictly need to initialize these with `null`, but seems like good practice? So at a glance one can see all the state that will be stored in this component.

src/renderer/InvestJob.js
dcdenu4: Ha, of course!

src/renderer/components/InvestTab/index.jsx
davemfish: Whew, that's a doozy of a question. This definitely deserves a code-comment. The idea here is to take the final line of stdErr and set it aside as "finalTraceback" for use in an `Alert`. The problem (I guess...it's been a while) is sometimes there are blank lines at the end of the stderr. So `splice(-1)` takes the last line in the array, but then keeps moving up from the back of the array by `i` until it finds a not blank line.

davemfish: I think this can probably become a whole lot more readable using `pop`. And in the process I'm making sure this is covered by a test. More soon.

davemfish:
> Hey @davemfish , just finished a pass at this. Only had a few comments / questions. Otherwise this looks like another iteration that is making things a bit nicer.

thanks @dcdenu4 ! I pushed a couple changes - one to rename some variables like you suggested, and one to simplify that really confusing stderr parsing stuff. I also added some more complex test data to better cover that parsing. Back to you!

PR #167, opened 2021-09-02
Bugfix/166 invest logging stdout

### LINKED ISSUE #166: Improve reliability of the connection to invest logging
The workbench displays invest logs by finding the invest logfile in the workspace and then reading/tailing it in the renderer to display the text. It seemed like a good idea to use the logfile (instead of the invest process stdout stream) as the single source of data for the logs because it lets us display the entire log from a pre-existing run in the same way we display it from a new, active run.

But it's tricky and a bit fragile to do this for new, active invest runs because it requires either,
1. identifying the logfile in the workspace using regex, and dealing with the reality of multiple logfiles in the same workspace, or
2. parsing the invest stdout itself for the logfile name, but then waiting until the file exists before trying to read/tail it

We've been doing option 1, and it's been fairly reliable, though I just had an issue with it that's hard to reproduce.

I think we can make this whole process more reliable & simpler by getting the log data from invest stdout during actively running jobs. And only getting the data from a logfile when we're rendering a pre-existing run. 

The `invest.cli` sets up a StreamHandler to `sys.stdout` and `invest.utils.prepare_workspace` sets up the FileHandler. So invest is already setup for this.

davemfish:
This fixes natcap/invest-workbench#166

We're splitting the display of invest logfiles into two distinct modes:
1. During an active invest run, the invest log is displayed by listening to the invest process stdout & stderr
2. To display the log of a pre-existing invest run loaded from the list of "Recent Runs", read the logfile from disk

We no longer need to worry about watching for logfiles to be created in active workspaces, or with watching the file itself for updates during an active run.

There's also some new tracking of the state of an invest subprocess so we can display the appropriate components. For example, the 'Cancel' button should only be available after we are certain the invest process is running and has a `pid`. But the 'Run' button should be disabled even earlier, as soon as it is clicked, and not enabled again until invest exits.

Also in this PR, a bit of debugging for the puppeteer test that should make it more resilient to things that seem to take varying amounts of time, like the electron startup, and invest validation.


src/main/main.js
```diff
@@ -106,21 +108,22 @@ export const createWindow = async () => {
     if (ELECTRON_DEV_MODE) {
       mainWindow.webContents.openDevTools();
     }
+    // We use this stdout as a signal in a puppeteer test
+    process.stdout.write('main window loaded');
```
phargogh, line 112: Neat!  Very helpful comment, too.

src/main/setupInvestHandlers.js
```diff
@@ -127,31 +90,38 @@ export function setupInvestRunHandlers(investExe) {
...
-      runningJobs[args.workspace_dir] = investRun.pid;
-      event.reply(`invest-logging-${channel}`, logfile);
-    });
+    // We need to store the PID to enable killing the task.
+    // And need to parse the logfile path from stdout.
+    const stdOutCallback = async (data) => {
+      if (!investLogfile) {
+        if (`${data}`.match('Writing log messages to')) {
+          investLogfile = `${data}`.split(' ').pop().trim();
+          runningJobs[args.workspace_dir] = investRun.pid;
```
phargogh, line 99: Out of curiosity, is it possible for a user to start 2 model runs (could be different models) in the same workspace with different PIDs?  If it is, wouldn't we lose track of the PID of the first process?

src/main/setupInvestHandlers.js
```diff
@@ -166,3 +136,25 @@ export function setupInvestRunHandlers(investExe) {
...
+      logger.error(err);
+      event.reply(
+        `invest-stdout-${channel}`,
+        `Logfile is missing or unreadable: ${os.EOL}${logfile}`
+      );
+    });
+
+    const rl = readline.createInterface({
+      input: fileStream,
+      crlfDelay: Infinity,
```
phargogh, line 153: Well this is an interesting option!  I wonder what their use case was for expecting there to be an observable delay between `\r` and `\n`.

src/main/setupInvestHandlers.js
```diff
@@ -166,3 +136,25 @@ export function setupInvestRunHandlers(investExe) {
...
+    const fileStream = fs.createReadStream(logfile);
+    fileStream.on('error', (err) => {
+      logger.error(err);
+      event.reply(
+        `invest-stdout-${channel}`,
+        `Logfile is missing or unreadable: ${os.EOL}${logfile}`
+      );
+    });
+
+    const rl = readline.createInterface({
```
phargogh, line 151: I'm sure this is just my relative node inexperience coming out, but I'm not sure I understand the benefit of having a readline interface atop `fileStream` ... is it just that we're trying to iterate one line at a time rather that whatever chunks `fileStream` would normally read?  Is there a limit to the size of the data that can be sent across the `invest-stdout-*` channel?  Something else?  Sorry for my confusion here!

phargogh (commented):
Thanks @davemfish !  I've looked through and had a couple questions for my own information that are very likely just due to my relative inexperience with node, but I'm curious what you think.

src/main/setupInvestHandlers.js
davemfish: Oh thanks for catching this! I switched the key to the `jobID` that is unique to each `InvestTab` component. Since we know we can only have one running job per open tab, that should fix this.

src/main/setupInvestHandlers.js
davemfish: I guess it must have happened at some point! Based on the docs `Infinity` seems like a reasonable cross-platform choice without any downside in our case.

src/main/setupInvestHandlers.js
davemfish: @phargogh That's a great question that I never thought too deeply about. The goal is to read one line at a time because that's how the `LogTab` `markupLine` is designed to work. If you don't mind, could you take a look in `LogTab` and give that design a bit of critical review? I wouldn't want to design this whole `fileStream + readline` stuff around something else that was a flawed idea to begin with.

At some point we'll probably move that `markupLine` feature out of the renderer and into this main process because it's still using one node module (`santizeHtml`) and we ought to eventually remove node integration from the renderer completely. That could mean just substituting in another `ipc` call.

src/main/setupInvestHandlers.js
phargogh: Ah right!  That makes a lot of sense.  Thanks!

src/main/setupInvestHandlers.js
phargogh: Sure!  So, I'm not certain I can speak to whether `markupLine` belongs in the main process or in the renderer at the moment, but your reasoning makes sense to me about that.

The main thing I'm noticing about this design centers around the line-by-line nature of it.  That's not _inherently_ a problem, but the fact that we start from the very beginning with reading from stdout line-by-line is probably less efficient than it could be.  The [Node docs for `createReadStream`](https://nodejs.org/api/fs.html#fs_fs_createreadstream_path_options) suggest that it'll by default read in 16Kb at a time, which is usually a whole lot more than a single line!  The point of deliberately passing around larger chunks is that it tends to reduce iteration overhead, and since each chunk here will be passing through the IPC (where presumably there's some overhead per call), I'd wonder about all of that overhead just generally slowing things down.

So if I were to make an architectural observation, it would be that we might be able to reduce overhead by reading data in larger chunks (the default should be fine) and then only split by newlines where it's needed so that we minimize that communication overhead.  It might not make a difference with just running 1 model, but if we're trying to monitor a bunch of streams all at the same time, I bet that would become noticeable.

And maybe this wouldn't matter at all if we move `markupLine` out of the renderer and we'll need more IPC for that move?  I'm not sure.

And I could of course be totally off base with this ... [my timings](https://gist.github.com/phargogh/00aea868fddcab3def00840da5482dcb) were around simply reading from a file stream, which is fairly simple compared with inter-process communication.  Anyways, I'd be curious what you think about this.

phargogh: We talked about this on the office hours huddle today and decided that there isn't enough of a slowdown to warrant a complete refactor of the line-by-line handling of markup, and we can always try to optimize it later on when we do have a clear slowdown.  @davemfish was also going to try using python's `PYTHONUNBUFFERED=1` environment variable to see how well that works and whether that might be useful and/or faster within the workbench context of reading from stdout.

davemfish: @phargogh here's an update. Adding the `PYTHONUNBUFFERED` env variable seems to work. There's no discernable difference with or without it for the invest runs I've tried, but this still seems like a good use-case for it.

However, reading line-by-line from a file on disk takes too long for even a modest-sized logfile of 1660 lines. It takes a few seconds to load. 

After more thinking, it might be reasonable to only do the markup during active invest runs as the data comes through stdout. Then for loading "Recent Run" logfiles, we just load it all as plaintext with no markup. In the future, if we want to add the markup feature to "Recent Run" logs, we might consider caching the marked-up logfile that we create in-memory during the invest run, and load *that* back later instead of the plaintext logfile. What do you think?

davemfish: Ditching line-by-line would also solve another minor problem: reading logfiles line-by-line is not exactly the same as logger message-by-message. As some messages contain newlines. So there were occasional cases where the pattern-matcher only marked-up the first line of a multi-line logger message. And our regex got more complicated in order to deal with that, especially for Tracebacks.

phargogh: > And our regex got more complicated in order to deal with that, especially for Tracebacks.

Yeah, I was wondering about this case as well.  Remembering back to that PyBay talk we had both attended, would be be worth considering structured logging?

> However, reading line-by-line from a file on disk takes too long for even a modest-sized logfile of 1660 lines. It takes a few seconds to load.
>
> After more thinking, it might be reasonable to only do the markup during active invest runs as the data comes through stdout. 

Oh my, that's a notable delay on a not-terribly long file!  I like your solution of marking up the file, caching it and presenting that to the user.  That feels like a good balance of orthogonality, keeping the UI features in the UI layer and still minimizing the need to recompute the HTML markup on file load.

davemfish:
Okay @phargogh I think this is ready for review again. I made #169 based on the outcome of our discussion. 

As for structured logging in invest vs. parsing plaintext with regex...I don't think we need to change invest logging for the benefit of the Workbench right now. Generally I think invest logs should stay nice & readable as text files and adding more structure to them might make them less readable.

src/main/setupInvestHandlers.js
```diff
@@ -112,50 +102,66 @@ export function setupInvestRunHandlers(investExe) {
...
-    // Capture stderr to a string separate from the invest log
-    // so that it can be displayed separately when invest exits.
-    // And because it could actually be stderr emitted from the
-    // invest CLI or even the shell, rather than the invest model,
-    // in which case it's useful to logger.debug too.
-    investRun.stderr.on('data', (data) => {
+    const stdErrCallback = (data) => {
+      // The python Traceback for invest comes through stdout & stderr,
+      // So no need to merge those streams for the benefit of displaying
+      // a complete log.
```
davemfish, line 148: This was an unexpected realization. It's convenient, but do we know exactly why this happens. And also there's at least one case where the Traceback is *not* going through stdout. https://github.com/natcap/invest/issues/651

phargogh: Based on what I found while working on https://github.com/natcap/invest/pull/655, the intent of the way InVEST's logging is structured in a `--headless` run is that logging will be sent to 2 handlers: `StreamHandler(sys.stdout)` and a `FileHandler`.  So a `LOGGER.exception` call should appear in both, and an uncaught exception (the CLI doesn't handle an exception in `execute`) will be printed to `stderr` as you'd expect for a python program.  It does seem like there's some undefined behavior around whether an exception message would be written to `stdout` via logging before the application terminates and the exception is written to `stderr`.

The changes in https://github.com/natcap/invest/pull/655 _should_ make exception handling more consistent and should clarify to python that we want to log the exception in the CLI if `execute` raises one.

phargogh (changes requested):
Makes sense to me!  I had a couple of very minor comments and a few questions, but nothing major I hope.

src/main/setupInvestHandlers.js
```diff
@@ -20,52 +20,42 @@ const LOGLEVELMAP = {
...
 
+const LOG_TEXT_TAG = 'span';
+const ALLOWED_HTML_OPTIONS = {
+  allowedTags: [LOG_TEXT_TAG],
+  allowedAttributes: { [LOG_TEXT_TAG]: ['class'] },
+};
+const LOG_ERROR_REGEX = /(Traceback)|(([A-Z]{1}[a-z]*){1,}Error)|(ERROR)/;
+export const LOG_PATTERNS = {
+  'invest-log-error': LOG_ERROR_REGEX,
+  'invest-log-primary': /a^/, // default is regex that will never match
```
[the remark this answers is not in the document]
davemfish: That's a good point. I changed this default value and the function to skip the pattern matching if the pattern is not a regex.

src/main/setupInvestHandlers.js
```diff
@@ -112,50 +102,66 @@ export function setupInvestRunHandlers(investExe) {
     ];
     logger.debug(`set to run ${cmdArgs}`);
     let investRun;
+    let investLogfile;
+    const env = {
+      PATH: path.dirname(investExe),
+      PYTHONUNBUFFERED: 'TRUE',
```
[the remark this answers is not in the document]
davemfish: I actually removed this env var since discovering that `logging` flushes with each emit. Since our goal is to get each logger message one at a time over stdout, I think that's all we need! Do you concur? https://github.com/python/cpython/blob/main/Lib/logging/__init__.py#L1092

src/main/setupInvestHandlers.js
```diff
@@ -112,50 +102,66 @@ export function setupInvestRunHandlers(investExe) {
...
+        }
+      }
+      // we set python stdio to be unbuffered, so data here should
+      // only be one logger message at a time.
+      event.reply(
+        `invest-stdout-${jobID}`,
+        markupMessage(`${data}`, LOG_PATTERNS)
+      );
+    };
+    LOG_PATTERNS['invest-log-primary'] = new RegExp(pyModuleName);
```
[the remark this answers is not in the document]
davemfish: Good catch, these would be in conflict. I fixed that bug.

src/renderer/components/InvestTab/index.jsx
```diff
@@ -110,18 +113,21 @@ export default class InvestTab extends React.Component {
...
-      this.switchTabs('log');
+      // Only now do we know for sure the process is running
+      updateJobProperties(jobID, {
+        logfile: logfile,
+        status: 'running',
+      });
     });
     ipcRenderer.on(`invest-stderr-${jobID}`, (event, data) => {
+      // It's convenient to have stderr in it's own object to display
+      // it an Alert.
```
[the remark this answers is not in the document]
davemfish: I moved this comment to somewhere where I think it's needed more.

src/main/setupInvestHandlers.js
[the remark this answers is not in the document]
phargogh: Yep, I concur!  That should do the trick for us here.

davemfish:
This should also fix #131 . Some of the trouble resolving that has been a bug I just discovered and fixed here:

We were registering `ipcRenderer.on` listeners every time the Run button is clicked. Which means it was easy to get duplicate listeners for the data being sent from `investRun.stderr.on('data', stdErrCallback)`, for example. So I think that was polluting the stderr strings that we trying to pattern-match on to filter out that `Failed to execute script ....` message from pyinstaller. I added to a test to cover filtering out that message.

phargogh (approved):
Thanks Dave!

PR #177, opened 2021-10-07
Task/51 usage logging

davemfish:
This PR adds invest usage logging to the Workbench. It relies on changes in https://github.com/natcap/invest/pull/678

Logging is done by `natcap.invest.usage` via a `ui_server` endpoint. The workbench requests the logging when the model starts and exits. The solution here is not nearly as elegant as the context manager we use to do this in Qt, but that's just not an option with node & spawned subprocesses.

This is setup to only do the logging in production mode, and not during puppeteer tests.

Database entries look something like this now:
```
| model_name                       | invest_release           | invest_interface      | time                     | ip_address | bounding_box_union                                                                 | bounding_box_intersection                                                          | node_hash | system_full_platform_string | system_preferred_encoding | system_default_language | session_id                           |
+----------------------------------+--------------------------+-----------------------+--------------------------+------------+------------------------------------------------------------------------------------+------------------------------------------------------------------------------------+-----------+-----------------------------+---------------------------+-------------------------+--------------------------------------+
| natcap.invest.carbon             | 3.9.1.post451+gc2a32f316 | Workbench 0.1.0-alpha | 2021-10-07T19:18:15.017Z | NULL       | [-123.70489878706152, 44.249954424247804, -123.20882889655645, 44.762121780513645] | [-123.70489878706152, 44.249954424247804, -123.20882889655645, 44.762121780513645] | NULL      | Windows-10-10.0.19043-SP0   | cp1252                    | en_US                   | 394571e3-6950-4538-b0a3-8b336c51d0c3 |
| natcap.invest.carbon             | 3.9.1.post446+gfb9442c9e.d20211005 | Qt               | 2021-10-05T20:00:04.929Z | NULL       | [-123.70488835731115, 44.24994809617993, -123.20882373475214, 44.7621148918994] | [-123.70488835731115, 44.24994809617993, -123.20882373475214, 44.7621148918994] | NULL      | Windows-10-10.0.19043-SP0   | cp1252                    | en_US                   | bccb8767-1000-4fd8-8b5d-71d2e055d562 |
```

This PR also upgrades to electron v15 to get a version of node 14.x that includes `crypto.randomUUID`. No changes were needed to support the upgrade.

There's also a little bit of cleanup in `InvestTab` as I noticed a slightly simpler way to manage invest stderr data.

### DIFF
--- a/src/main/isDevMode.js
+++ b/src/main/isDevMode.js
@@ -1,0 +1,4 @@
+// defaultApp property is added by electron in devmode.
+// It means the app is running in the "default" electron app
+// as opposed to a purposely-built package, as in production.
+export default !!process.defaultApp;
--- a/src/main/setupInvestHandlers.js
+++ b/src/main/setupInvestHandlers.js
@@ -10,9 +10,11 @@ import sanitizeHtml from 'sanitize-html';
 
 import { getLogger } from '../logger';
 import { ipcMainChannels } from './ipcMainChannels';
+import pkg from '../../package.json';
 
 const logger = getLogger(__filename.split('/').slice(-1)[0]);
 
+const WORKBENCH_VERSION = pkg.version;
 // to translate to the invest CLI's verbosity flag:
 const LOGLEVELMAP = {
   DEBUG: '--debug',
@@ -145,7 +147,7 @@ export function setupInvestRunHandlers(investExe) {
               body: JSON.stringify({
                 model_pyname: pyModuleName,
                 model_args: JSON.stringify(args),
-                invest_interface: 'Workbench', // TODO: include wb version
+                invest_interface: `Workbench ${WORKBENCH_VERSION}`,
                 session_id: usageSessionId,
               }),
               headers: { 'Content-Type': 'application/json' },
--- a/src/main/setupInvestHandlers.js
+++ b/src/main/setupInvestHandlers.js
@@ -74,7 +74,7 @@ export function setupInvestRunHandlers(investExe) {
     event, modelRunName, pyModuleName, args, loggingLevel, jobID
   ) => {
     let investRun;
-    let investLogfile;
+    let investStarted = false;
     let investStdErr = '';
     const logPatterns = { ...LOG_PATTERNS };
     logPatterns['invest-log-primary'] = new RegExp(pyModuleName);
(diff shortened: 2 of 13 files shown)


src/main/isDevMode.js
```diff
@@ -0,0 +1,4 @@
+// defaultApp property is added by electron in devmode.
+// It means the app is running in the "default" electron app
+// as opposed to a purposely-built package, as in production.
+export default !!process.defaultApp;
```
dcdenu4, line 4: What's the double `!!` signify? I don't think I've seen that yet.

src/main/setupInvestHandlers.js
```diff
@@ -116,23 +123,27 @@ export function setupInvestRunHandlers(investExe) {
         detached: true, // counter-intuitive, but w/ true: invest terminates when this shell terminates
       });
     } else { // windows
-      investRun = spawn(investExe, cmdArgs, {
+      investRun = spawn(`"${investExe}"`, cmdArgs, {
```
dcdenu4, line 126: Is this to help protect against spaces in the path?

src/main/setupInvestHandlers.js
```diff
@@ -156,14 +167,20 @@ export function setupInvestRunHandlers(investExe) {
 
     investRun.on('exit', (code) => {
       delete runningJobs[jobID];
-      event.reply(`invest-exit-${jobID}`, code);
+      event.reply(`invest-exit-${jobID}`, {
+        code: code,
+        stdErr: investStdErr,
```
dcdenu4, line 172: Okay, so we can just bundle that stderr reply in here instead of having it be a separate call in the Callback itself.

tests/binary_tests/puppet.test.js
```diff
@@ -86,7 +86,10 @@ beforeAll(() => {
     `"${BINARY_PATH}"`,
     // these are chromium args
     [`--remote-debugging-port=${PORT}`],
-    { shell: true }
+    {
+      shell: true,
+      env: { ...process.env, PUPPETEER: true }
```
dcdenu4, line 91: What does the `PUPPETEER` env do?

src/main/isDevMode.js
davemfish: It's a conventional way to get a pure true false out of a variable that may or may not be `undefined`. Try it out!

src/main/setupInvestHandlers.js
davemfish: Yes, I noticed there was no need to send each stderr message individually to the renderer as they stream in. So we cut down on some ipc traffic by just sending the complete stderr at the end.

PR #181, opened 2021-10-18
Task/123 disable devtools & inspect element in production

### LINKED ISSUE #123: disable developer tools menu in production
By default there is `View > toggle developer tools` but this might be confusing and I can't think of reason why users would ever benefit from opening devtools. We should remove this in production. Users in-the-know could still open via the keyboard shortcut, and I think that's okay. It might be okay to remove that menu item from dev mode as well because devtools still open on launch and can be re-opened with the keyboard shortcut.

Also remove the right-click "Inspect Element" menu in production, but keep for dev mode.

davemfish:
This PR fixes #123 by removing the option to "Inspect Element" from the context menu and to "toggle developer tools" from the View menu. These options are now available in Dev mode but not in production.

There are also additions to the right-lick context menu to offer some standard options depending on what was clicked (cut/copy/paste/select all, etc). And now we're using the electron `webContents.on('context-menu)` API instead of `window.addEventListener('contextmenu')`, so the whole thing is setup in the main process instead of the renderer.

### DIFF
--- a/src/preload.js
+++ b/src/preload.js
@@ -6,7 +6,6 @@ const { getLogger } = require('./logger');
 
 function preload() {
   window.Workbench = {
-    isDevMode: !!process.defaultApp,
     getLogger: getLogger,
   };
 }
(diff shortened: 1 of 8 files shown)


src/preload.js
```diff
@@ -6,7 +6,6 @@ const { getLogger } = require('./logger');
 
 function preload() {
   window.Workbench = {
-    isDevMode: !!process.defaultApp,
```
davemfish, line 9: This never actually worked. Since this file runs in a renderer process it never has `process.defaultApp` defined. Also at this point nothing in the renderer needs to be aware of dev vs production modes.

PR #183, opened 2021-10-27
Code signing for mac

### LINKED ISSUE #28: Add code-signing to the build process

emlys:
Fixes #28.

Note that code signing doesn't run in PRs, so be sure to check the GHA run on my branch instead: https://github.com/emlys/invest-workbench/tree/task/codesigning

* Use the new Apple Developer ID Application certificate for mac signing
* Use the same old certificate for windows signing
* Sign as a separate step outside of electron-builder, because it was too hard to configure.

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 defaults:
   run:
-    shell: bash -l {0}
+    shell: bash -l -x {0}
 
 jobs:
   build-and-release:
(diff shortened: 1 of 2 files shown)


davemfish (approved):
Thanks, @emlys this looks good to me. I looked at the Actions on your branch and this looks like it's doing the job.

It doesn't seem to have any impact on the user-experience when launching the app the first time. But maybe we're not expecting that to improve at all until after we submit to apple for notarization.

PR #185, opened 2021-10-28
Bugfix/tests data

davemfish:
Hey @emlys , this is something minor we missed when updating tests in #178 . We had some test data (a sample `ARGS_SPEC`) living in a json file that we forgot to update with the `module` -> `pyname` change. I took this opportunity to remove that file and keep the test data in the script itself.

And then it turned out we no longer need a `tests/data` directory tracked in this repo. So I updated all the tests that use that as a scratch workspace to instead use `os.tmpdir()`.

I also turned back on our jest override that throws errors on warnings. That would have caught this sooner. Please see the lengthy comments in that file about why I turned it off at some point in the past.


tests/invest/flaskapp.test.js
```diff
@@ -39,140 +43,153 @@ afterAll(async () => {
...
-    model_module: spec.pyname,
-    args: JSON.stringify(argsDict),
-  };
-
-  const results = await server_requests.fetchValidation(payload);
-  // There's always an array of arrays, where each child array has
-  // two elements: 1) an array of invest arg keys, 2) string message
-  expect(results[0]).toHaveLength(2);
-});
+  test('invest list items have expected properties', async () => {
```
davemfish, line 56: Hmm the line-by-line diff of this file is not so useful because most of the changes are just moving blocks of code around, moving tests into a `describe` block.

emlys (changes requested):
@davemfish, this all looks good to me except I'm kind of confused about your comment about going back on the jest override... I'm not sure what that's referring to here. Is there maybe a change missing in `jest.overrides.js`?

davemfish:
@emlys the macos pptr test is failing for an interesting reason. The workbench logs show:

```
[15:21:16.093] [main.js] natcap/invest/validation.py:692: UserWarning: Validation of file /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T timed out. If this file is stored in a file streaming service, it may be taking a long time to download. Try storing it locally instead.
```

So maybe the change in this PR to use `os.tmpdir()` as the workspace is causing this. But I wouldn't expect that folder to take any longer to read from than any other. I'll look into this further.

davemfish:
Okay @emlys I'm done messing around with this and it's ready to review again. Thanks!

emlys (approved):
Looks good @davemfish!

PR #186, opened 2021-11-01
Task/170 replace material designs icons

### LINKED ISSUE #170: material-design-icons failures on yarn install in GHA
I believe this is the same issue as our past intermittent failures on `npm install` in GHA that inspired the switch to yarn. This is happening less often with yarn, but there's still occasionally a failure and I finally got some verbose logging from it:

```
verbose 198.444199837 Error: unexpected end of file
    at Zlib.zlibOnError [as onerror] (zlib.js:188:17)
error An unexpected error occurred: "https://registry.yarnpkg.com/material-design-icons/-/material-design-icons-3.0.1.tgz: unexpected end of file".
```

If it ever happens again, will be interesting to note if it's always `material-design-icons`. It's by far the biggest package we're installing. Here's the top 5 biggest in my local `node_modules`

```
356M	node_modules/material-design-icons
194M	node_modules/electron
178M	node_modules/react-devtools
105M	node_modules/app-builder-bin
12M	node_modules/@jest
```

davemfish:
This fixes #170 . I took Doug's advice and switched to `react-icons`, which it turns out still provides all the material designs icons, but in a much smaller package! They're easier to use also since they are react components, so I used them in a variety of places where we were not using icons at all.

Test queries needed updating in a lot of cases where I removed text from the DOM. Whenever button text was completely replaced with an icon, I added an `aria-label` to keep the button accessible to screen-readers and to `findByRole` test queries. And then it also turned out that those new aria labels often include the name of an invest arg, which means there are now multiple DOM elements with labels including the same text. Hence the need to make queries more specific, either by removing `RegExp` wrappers (not sure why we ever needed/wanted those) or switching to `findByRole`.


src/renderer/components/ResourcesLinks/index.jsx
```diff
@@ -60,10 +62,12 @@ export default function ResourcesTab(props) {
...
     <React.Fragment>
       <a href={userGuideURL} onClick={handleClick}>
-        {"User's Guide"}
+        <MdOpenInNew />
+        {" User's Guide"}
       </a>
       <a href={forumURL} onClick={handleClick}>
-        Frequently Asked Questions
+        <MdOpenInNew />
+        {' Frequently Asked Questions'}
```
dcdenu4, line 70: Are the JSX brackets useful here to get that space in the string? Instead of using a space as a separator between the icon and text, I wonder if it'd be a better practice to add a CSS class to icons and use `margin-right` or `pad`?

src/renderer/components/SaveFileButton/index.jsx
```diff
@@ -32,7 +33,8 @@ export default class SaveFileButton extends React.Component {
         onClick={this.browseSaveFile}
         variant="link"
       >
-        {this.props.title}
+        <MdSave />
+        {` ${this.props.title}`}
```
dcdenu4, line 37: Same question about the best practice for spacing between icon and text.

tests/renderer/app.test.js
```diff
@@ -226,32 +229,32 @@ describe('Various ways to open and close InVEST models', () => {
 
     // Click the close button on the middle tab
     const tab2CloseButton = await within(tab2.closest('.nav-item'))
-      .getByRole('button', { name: /x/ });
+      .getByRole('button', { name: new RegExp(`close ${MOCK_MODEL_TITLE}`) });
```
dcdenu4, line 232: Nice, using that arial-label.

tests/renderer/app.test.js
```diff
@@ -339,87 +342,124 @@ describe('Display recently executed InVEST jobs on Home tab', () => {
...
 
     const {
-      getByText, getByLabelText, getByTitle, findByTitle
+      getByText, getByLabelText, findByRole
     } = render(
       <App />
     );
 
-    fireEvent.click(await findByTitle('settings'));
+    userEvent.click(await findByRole('button', { name: 'settings' }));
```
dcdenu4, line 382: Out of curiosity, was there a particular reason you switched to `userEvent`? I wish the docs were more helpful in establishing the differences between the two.
>Most projects have a few use cases for fireEvent, but the majority of the time you should probably use @testing-library/user-event.

tests/renderer/setuptab.test.js
```diff
@@ -153,33 +160,36 @@ describe('Arguments form interactions', () => {
...
-    expect(await findByText('Browse')).toBeInTheDocument();
+    expect(await findByRole('button', { name: /browse for/ }))
+      .toBeInTheDocument();
 
     // Browsing for a file
     const filepath = 'grilled_cheese.csv';
     let mockDialogData = { filePaths: [filepath] };
     ipcRenderer.invoke.mockResolvedValue(mockDialogData);
-    fireEvent.click(await findByText('Browse'));
+    fireEvent.click(await findByRole('button', { name: /browse for/ }));
```
dcdenu4, line 184: Is there a reason to keep `fireEvent` here and not user `userEvent`?

dcdenu4 (changes requested):
Hey @davemfish I really like the new look of UI with the icons. I just had one suggestion and one question! Let me know what you think.

tests/renderer/app.test.js
davemfish: The difference as I understand it (don't remember where I read it) is fireEvent fires the single event (e.g. "click") and userEvent mimics the actual user interactions (e.g. mousedown, mouseup) which result in the click event.

Basically this test was giving me issues for other reasons and it was kind of big and unwieldy, so I split it up a bit and made this update in the process.

tests/renderer/setuptab.test.js
davemfish: Nope, no reason. FireEvent is still used in tons places in our tests. I only switched to userEvent for that one test that I was touching a bit more heavily anyway. 

If we think it's worth doing, I'd say we make the switch everywhere in a dedicated PR.

davemfish:
> Hey @davemfish I really like the new look of UI with the icons. I just had one suggestion and one question! Let me know what you think.

Yeah the icons really tie the room together, don't you think? Good advice to use a class instead of a mysterious whitespace character next to those icons. I made that change. Back to you @dcdenu4 !

src/renderer/components/SetupTab/index.jsx
```diff
@@ -401,7 +402,8 @@ export default class SetupTab extends React.Component {
                 onClick={this.browseForDatastack}
                 variant="link"
               >
-                Load parameters from file
+                <MdFolderOpen />
+                {' Load parameters from file'}
```
dcdenu4, line 406: I think this is supposed to be replaced by a class too?

src/renderer/components/ResourcesLinks/index.jsx
```diff
@@ -60,9 +62,11 @@ export default function ResourcesTab(props) {
   return (
     <React.Fragment>
       <a href={userGuideURL} onClick={handleClick}>
-        {"User's Guide"}
+        <MdOpenInNew className="mr-1" />
```
dcdenu4, line 65: I didn't see `mr-1` added to the CSS. Is this defined somewhere else?

dcdenu4 (changes requested):
Hey @davemfish , I think I found one missed spot for replacing the space with a className. I also was wondering where the `mr-1` CSS was defined. I didn't see it in `style.css`

src/renderer/components/SetupTab/index.jsx
davemfish: Whoops, thank you!

src/renderer/components/ResourcesLinks/index.jsx
davemfish: It's defined for us in bootstrap

davemfish:
> Hey @davemfish , I think I found one missed spot for replacing the space with a className. I also was wondering where the `mr-1` CSS was defined. I didn't see it in `style.css`

Thanks again! Change pushed. 

It's a bootstrap class 👍

dcdenu4 (approved):
Thanks Dave!

PR #188, opened 2021-11-04
Task/121 model list order

### LINKED ISSUE #121: Order the invest model list in a smart way
Now, we're defaulting to the order returned by `invest list`, which is alphabetical, but might not actually alphabetize on the same form of the name that is displayed on the button.

Anyway, I think we can do better than alphabetizing. Some options to consider:

* Keep the user's frequently used models in a section at the top, alpha-sort for everything else
* Use usage logging data, or forum post data, to order the models by general popularity
* Sort models into sections by domain (Urban, Freshwater, Coastal, etc) This will certainly breakdown as some models are hard to classify. And those categories might mean a lot less to users than they do to us insiders. 
* Use the same method as the User's Guide - sections for Supporting, Final, Urban ES, & other tools. Personally, I don't find those headings to be useful.
* Alphabetize based on the same names that are displayed on the buttons.

Always keep CBC Preprocessor & CBC together, in that order.

### LINKED ISSUE #184: Cancel Run button depends on the python loggingLevel
The Cancel button is only available after we know that the invest run has started. And we rely on listening for certain stdout from invest to determine that. Apparently we're listening for something that can be filtered out by setting a loggingLevel at `ERROR`, for example.

I think there's already an invest logging message that always logs with level `100`, so we should probably look for that message instead. Or create a new, dedicated signal for this purpose.

davemfish:
This PR fixes #121. It orders the model list alphabetically by the model title text. It always keeps CBC Precprocessor before CBC model. 

It also wraps the list into two columns so that all the buttons are visible onscreen without scrolling.

Bonus: I updated the invest version to the HEAD of `release/3.10` , which is all we need to fix #184

### DIFF
--- a/src/styles/style.css
+++ b/src/styles/style.css
@@ -133,21 +133,27 @@ body {
 
 /* Home Tab */
 
-.invest-list-group {
-	display: flex;
-	flex-direction: column;
-	flex-wrap: wrap;
+.invest-list-container {
 	height: 87vh;
 	overflow-y: auto;
 }
 
+.invest-list-group {
+	/*do not constrain height here because it can cause more
+	columns to be created. Constrain it in parent instead.*/
+	display: block;
+	column-count: 2;
+	white-space: nowrap;
+}
+
 .invest-list-group .list-group-item {
-	width: auto;
 	border-top: none;
 	border-right: none;
 	border-left: none;
 	border-bottom: 1px solid rgba(0,0,0,.125);
 	margin-bottom: 0;
+	overflow: hidden;
+	text-overflow: ellipsis;
 }
 
 .invest-list-group .list-group-item:last-child {
--- a/src/styles/style.css
+++ b/src/styles/style.css
@@ -1,6 +1,7 @@
 body {
 	--invest-green: #148F68;
 	background-color: rgba(0,0,0,0);
+	font-size: 16px;
 	margin: 0;
 	padding: 1rem;
 	overflow-y: hidden;
--- a/src/styles/style.css
+++ b/src/styles/style.css
@@ -132,19 +132,36 @@ body {
 
 /* Home Tab */
 
-.invest-list-table {
-	display: block;
+.invest-list-group {
+	display: flex;
+	flex-direction: column;
+	flex-wrap: wrap;
 	height: 87vh;
 	overflow-y: auto;
 }
 
-.invest-list-table .table td {
+.invest-list-group .list-group-item {
+	width: auto;
 	border-top: none;
-	border-bottom: 1px solid #dee2e6;
+	border-right: none;
+	border-left: none;
+	border-bottom: 1px solid rgba(0,0,0,.125);
+	margin-bottom: 0;
+}
+
+.invest-list-group .list-group-item:last-child {
+	border-radius: 0;
 }
 
 .invest-button {
 	color: var(--invest-green);
+	font-weight: 600;
+}
+
+.invest-button:hover {
+	color: white;
+	background-color: var(--invest-green);
+	opacity: 75%;
 	font-weight: 600;
 }
 
--- a/src/styles/style.css
+++ b/src/styles/style.css
@@ -1,7 +1,10 @@
+:root {
+	font-size: 16px;
+}
+
 body {
 	--invest-green: #148F68;
 	background-color: rgba(0,0,0,0);
-	font-size: 16px;
 	margin: 0;
 	padding: 1rem;
 	overflow-y: hidden;
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
   "invest": {
     "bucket": "releases.naturalcapitalproject.org",
     "fork": "natcap",
-    "version": "3.9.1.post531+g2159bd07",
+    "version": "3.9.2.post470+g11795c5c",
     "target": {
       "macos": "mac_invest_binaries.zip",
       "windows": "windows_invest_binaries.zip"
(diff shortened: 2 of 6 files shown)


src/styles/style.css
```diff
@@ -1,6 +1,7 @@
 body {
 	--invest-green: #148F68;
 	background-color: rgba(0,0,0,0);
+	font-size: 16px;
```
emlys, line 4: Should this be in `rem` to allow it to scale?

src/styles/style.css
```diff
@@ -132,22 +133,39 @@ body {
...
 }
 
-.invest-list-table .table td {
+.invest-list-group .list-group-item {
+	width: auto;
 	border-top: none;
-	border-bottom: 1px solid #dee2e6;
+	border-right: none;
+	border-left: none;
+	border-bottom: 1px solid rgba(0,0,0,.125);
```
emlys, line 149: Just out of curiosity, are RGBA colors preferred over hex?

src/styles/style.css
emlys: I would think 1 rem so it starts at the default size

src/styles/style.css
```diff
@@ -132,22 +133,39 @@ body {
 
 /* Home Tab */
 
-.invest-list-table {
-	display: block;
+.invest-list-group {
+	display: flex;
+	flex-direction: column;
+	flex-wrap: wrap;
```
emlys, line 139: So this will wrap to make as many columns as needed to fit them all on the screen (which is currently 2 columns)? Do more columns form if you change the window size?

emlys (changes requested):
@davemfish just had a couple questions and a suggestion about `rem`!

src/styles/style.css
davemfish: So the only reason I added this was because I was asking myself, "what controls the default font size?". I learned the browser has a default set at 16px, but thought I would define it here in case I'm asking that question again in 6 months.

I guess it could be `1rem`, but it's functionally no different unless chromium changes it's underlying default size. For all other rules except this "root" font-size rule, I agree we should use `rem`. 

The font is still scalable by the user this way, and the browser still remembers if you increased/decreased the size last time you had it open.

src/styles/style.css
davemfish: I really don't know. The diff makes it looks like that was the main point of this change, but that's just a coincidence. I had purged all the old rules before creating the new ones.

src/styles/style.css
davemfish: Yes, it wraps to a second column if the list is too long vertically (which it always will be). More columns appear if the window height shrinks too much for two columns. And a horizontal scroll appears if window width shrinks too much.

But my aim was for no scrollbars and two columns by default. So the aim is for that to hold true for most screen sizes with the default window & font sizes.

davemfish:
Thanks @emlys , back to you! I didn't make any changes, but don't feel too strongly about that `rem` vs `px` one.

davemfish:
@emlys I tried this on an old-ish laptop and the layout here did not neatly fit in two columns with no scrollbar. So I'm going to try something different to achieve that. Putting in draft mode for now.

davemfish:
Here's how this looks now on my laptop which is only 1280 x 800. Considering most screens will have more modern pixels than this, I think this is a good default layout.

workbench

I didn't try to adjust the window size for the puppeteer tests because it would actually require changing the virtual screen size in the VM. Technically possible, but probably depends on the OS. And there is other weirdness at play too, like scrollbars never show in those screenshots even at window sizes where they definitely exist. It's just not worth trying to use GHA to get really realistic screenshots, at this point, I don't think.

src/styles/style.css
emlys: Okay, sounds good. I think if we do want to set the default font size, the most correct way to do that would be with [`:root`](https://developer.mozilla.org/en-US/docs/Web/CSS/:root), because `rem`s are based off the root element, and `<body />` is not the root element. But in practice it won't make a difference - they should always be the same.

emlys (dismissed):
Looks good to me! Approving as-is, also left another comment about the `rem` thing.

>  change from flex to block column layout to enforce two columns and vertical scroll if necessary.

Honestly I'd prefer using just one solution to the model list overflow - multiple columns, or vertical scroll, but not both. But this does accomplish the goal of showing most of the model names on most screens/window sizes.

davemfish:
> Honestly I'd prefer using just one solution to the model list overflow - multiple columns, or vertical scroll, but not both. But this does accomplish the goal of showing most of the model names on most screens/window sizes.

I think I see what you mean. Now that you've planted the seed in my brain, it looks ever-so-slightly off to see that scrollbar with two columns. But yeah, I'm hoping it will be rare to see the scrollbar. And the two-columns makes better use of space than one column, I think.

davemfish:
@emlys I made that switch from `body` to `:root` that you suggested. So now I guess you need to re-approve. Thanks again!

src/styles/style.css
dcdenu4: This made me wonder about representing alpha (transparency) in hex, and sure enough you can tack on a leading two element marker to represent transparency. Fun!
<html>
<body>


Alpha Level | HEX Code
-- | --
100% | FF
95% | F2
90% | E6
85% | D9
80% | CC
75% | BF
70% | B3
65% | A6
60% | 99
55% | 8C
50% | 80
45% | 73
40% | 66
35% | 59
30% | 4D
25% | 40
20% | 33
15% | 26
10% | 1A
5% | 0D
0% | 00


</body>
</html>

emlys (approved):
Thanks @davemfish!

PR #190, opened 2021-11-08
replacing event.target with event.currentTarget in event handlers

### LINKED ISSUE #189: event listener callbacks should refer to event.currentTarget
In some places, we're referring to `event.target` where we should be using `event.currentTarget`. 

Here's the difference:
> target = element that triggered event.
currentTarget = element that has the event listener.

https://stackoverflow.com/questions/10086427/what-is-the-exact-difference-between-currenttarget-property-and-target-property

This came up after replacing some button's text with svg icons. Now it's possible to click the svg, and while the button's event listener is still triggered, the `event.target` referenced in the handler function will be missing properties like `name` or `value` that we added to the button, not the svg.

davemfish:
This fixes #189 

`event.currentTarget` seems like it's always the right choice when the event handler function is attached to the same node that has attributes/properties we want to pass to the function. If we were registering the handler on a parent node where it could be shared among children, then we might have a use-case for `event.target`, but that's not a pattern we've been using thus far.

### DIFF
--- a/tests/renderer/setuptab.test.js
+++ b/tests/renderer/setuptab.test.js
@@ -195,6 +195,23 @@ describe('Arguments form interactions', () => {
     });
   });
 
+  test('Browse button populates an input - test click on child svg', async () => {
+    spec.args.arg.type = 'csv';
+    const {
+      findByRole, findByLabelText
+    } = renderSetupFromSpec(spec, uiSpec);
+
+    const filepath = 'grilled_cheese.csv';
+    const mockDialogData = { filePaths: [filepath] };
+    ipcRenderer.invoke.mockResolvedValue(mockDialogData);
+    const btn = await findByRole('button', { name: /browse for/ });
+    // Click on a target element nested within the button to make
+    // sure the handler still works correctly.
+    fireEvent.click(btn.querySelector('svg'));
+    expect(await findByLabelText(`${spec.args.arg.name}`))
+      .toHaveValue(filepath);
+  });
+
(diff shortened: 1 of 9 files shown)


tests/renderer/setuptab.test.js
```diff
@@ -195,6 +195,23 @@ describe('Arguments form interactions', () => {
     });
   });
 
+  test('Browse button populates an input - test click on child svg', async () => {
+    spec.args.arg.type = 'csv';
+    const {
+      findByRole, findByLabelText
+    } = renderSetupFromSpec(spec, uiSpec);
+
+    const filepath = 'grilled_cheese.csv';
```
dcdenu4, line 204: yum

tests/renderer/setuptab.test.js
```diff
@@ -195,6 +195,23 @@ describe('Arguments form interactions', () => {
...
+      findByRole, findByLabelText
+    } = renderSetupFromSpec(spec, uiSpec);
+
+    const filepath = 'grilled_cheese.csv';
+    const mockDialogData = { filePaths: [filepath] };
+    ipcRenderer.invoke.mockResolvedValue(mockDialogData);
+    const btn = await findByRole('button', { name: /browse for/ });
+    // Click on a target element nested within the button to make
+    // sure the handler still works correctly.
+    fireEvent.click(btn.querySelector('svg'));
```
dcdenu4, line 210: nice

dcdenu4 (approved):
Hey @davemfish this makes sense to me. A nice catch in subtlety.

PR #194, opened 2021-11-15
Task/187 package updates

### LINKED ISSUE #187: address dependabot security alerts
At or before the beta-release, we should clear the queue of suggested security updates. And also figure out the right way to manage this. The alerts are nearly always about packages that we don't directly specify in `package.json`. 

So is the correct approach to directly edit the `yarn.lock` file? We never need to manually touch that file for any other reason, and that seems like a good thing. Before resorting to that, we could try updating everything listed in `yarn outdated` and see if that addresses everything.

### LINKED ISSUE #856: address dependabot security alerts
At or before the beta-release, we should clear the queue of suggested security updates. And also figure out the right way to manage this. The alerts are nearly always about packages that we don't directly specify in `package.json`. 

So is the correct approach to directly edit the `yarn.lock` file? We never need to manually touch that file for any other reason, and that seems like a good thing. Before resorting to that, we could try updating everything listed in `yarn outdated` and see if that addresses everything.

davemfish:
This PR has the result of a `yarn upgrade`. It upgrades all dependencies to the version allowed by the `package.json`, which in our case always uses the `^` syntax to basically indicate bugfix and minor version upgrades are allowed, major version upgrades are not.

That upgrade should resolve the only open security alert, since we now have `ansi-regex v5.0.1`. Fixes natcap/invest#856 

Also moved one other dependency to the devDependencies section where it belongs. 

And also checked in a `HISTORY.rst` file since I didn't want to do separate PR for that :)

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,0 +1,10 @@
+.. :changelog:
+
+
+..
+  Unreleased Changes
+  ------------------
+
+0.1.0-beta (XXXX-XX-XX)
+-----------------------
+* First public release!
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
   "license": "MIT",
   "dependencies": {
     "bootstrap": "4.3.1",
-    "cross-env": "^5.2.1",
     "electron-log": "^4.3.5",
     "esm": "^3.2.25",
     "glob": "^7.1.6",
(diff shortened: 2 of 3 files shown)


phargogh (approved):
Makes sense to me!

PR #197, opened 2021-11-16
Bugfixes for sample data download issues

### LINKED ISSUE #195: sample data download complete alert - bad layout
A picture says it all...
download-alert

This alert should display in the same location as the progress bar it's replacing: to the left of the settings icon instead of on top of it.

### LINKED ISSUE #196: various sample data download issues
1. We have some names that are outdated, for example, we have sample data listed for "Hydropower", but we've switched to displaying "Annual Water Yield" everywhere else. Names in the download modal should all match model titles on the Home tab.
2. bucket URLs to zip files are broken if the zip has been replaced since it's original upload. We don't need to use the url that includes the `generation` parameter.
3. Stormwater model is missing from the sample data registry.

davemfish:
Fixes #195 and fixes #196 . Please see the descriptions on those issues.

### DIFF
--- a/scripts/fetch_invest_binaries.js
+++ b/scripts/fetch_invest_binaries.js
@@ -158,7 +158,7 @@ async function updateSampledataRegistry() {
   }
   fs.writeFileSync(
     DATA_REGISTRY_SRC_PATH,
-    JSON.stringify(registry, null, 2)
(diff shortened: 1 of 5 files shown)


dcdenu4 (approved):
Looks good to me @davemfish . I tested out downloading some data and it displays as desired here.

PR #199, opened 2021-11-17
Bugfix/198 save python script

### LINKED ISSUE #198: save to python script does not work
Here's the exception

```
58:07  natcap.invest.ui_server ERROR    Exception on /save_to_python [POST]
Traceback (most recent call last):
  File "flask\app.py", line 2073, in wsgi_app
  File "flask\app.py", line 1518, in full_dispatch_request
  File "flask\app.py", line 1516, in full_dispatch_request
  File "flask\app.py", line 1502, in dispatch_request
  File "natcap\invest\ui_server.py", line 196, in save_to_python
  File "natcap\invest\cli.py", line 150, in export_to_python
KeyError: 'Urban Cooling'
```

It looks like we're passing the wrong form of the model name.

davemfish:
Fixes #198 

We were passing the model title (e.g. Carbon Storage and Sequestration) instead of the model name (e.g. carbon). I updated tests to catch that if it ever happens again. 

We were also passing the model's module name to this endpoint, but that's not being used at all so I also removed it from the payload.

### DIFF
--- a/src/renderer/components/InvestTab/index.jsx
+++ b/src/renderer/components/InvestTab/index.jsx
@@ -285,7 +285,7 @@ export default class InvestTab extends React.Component {
               >
                 <SetupTab
                   pyModuleName={modelSpec.pyname}
-                  modelName={modelSpec.model_name}
+                  modelName={modelRunName}
                   argsSpec={argsSpec}
                   uiSpec={uiSpec}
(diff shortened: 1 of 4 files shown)


dcdenu4 (approved):
@davemfish looks good to me!

PR #201, opened 2021-11-18
Feature/200 settings dialog ux

### LINKED ISSUE #200: settings dialog is unintuitive
There are misc small things I'd like to improve about this UI:

Capture

A couple of the buttons have dedicated labels next to the button. We might as well put the label text on the button itself and remove some of the redundancy and clutter.

The Cancel & Save buttons only apply to the top section settings, so should be moved up to that section. But as @emlys suggested, it might be better to just automatically save on every change and do away with the save button. The Reset button then should also trigger a save. And Cancel could be replaced with an `X` in the upper right of the dialog.

davemfish:
Fixes #200 

In addition to the issues mentioned in the issue, this PR also switches the `n_workers` input field to a dropdown menu populated with numbers from -1 to the number of available CPUs. That data comes from NodeJS and is sent to the renderer at app startup. That removes the need to validate that input field.

I also added some expandable/collapsible helptext for that parameter, similar to what we have in the Qt UI. 

settings_ui

### DIFF
--- a/src/renderer/components/SettingsModal/index.jsx
+++ b/src/renderer/components/SettingsModal/index.jsx
@@ -1,12 +1,17 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import Accordion from 'react-bootstrap/Accordion';
 import Col from 'react-bootstrap/Col';
 import Row from 'react-bootstrap/Row';
 import Form from 'react-bootstrap/Form';
 import Button from 'react-bootstrap/Button';
 import Modal from 'react-bootstrap/Modal';
-import { MdSettings, MdClose } from 'react-icons/md';
+import {
+  MdSettings,
+  MdClose,
+} from 'react-icons/md';
+import { BsChevronExpand } from 'react-icons/bs';
 
 import { getDefaultSettings } from './SettingsStorage';
 
@@ -92,8 +97,8 @@ export default class SettingsModal extends React.Component {
           </Modal.Header>
           <Modal.Body>
             <Form.Group as={Row}>
-              <Form.Label column sm="8" htmlFor="logging-select">Logging threshold</Form.Label>
-              <Col sm="4">
+              <Form.Label column sm="7" htmlFor="logging-select">Logging threshold</Form.Label>
+              <Col sm="5">
                 <Form.Control
                   id="logging-select"
                   as="select"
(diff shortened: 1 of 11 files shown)


davemfish:
Sorry for the future merge conflicts @emlys since you already changed some of this stuff in your branch!

src/renderer/components/SettingsModal/index.jsx
```diff
@@ -1,98 +1,68 @@
...
 export default class SettingsModal extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       show: false,
-      localSettings: {
-        nWorkers: '',
-        loggingLevel: '',
-        sampleDataDir: null,
-      },
```
emlys, line 34: I like that we're no longer having a separate copy of the settings in this component state!

emlys (changes requested):
@davemfish, I like the dropdown for the n workers value! To further streamline it, instead of having the options '-1' and '0' in the dropdown, then defining them in the collapsible box, what if the definitions were right in the dropdown? So the options would be like "Synchronous (recommended)", "Multithreaded single process", "2 processes", "3 processes", ...

emlys:
> Sorry for the future merge conflicts @emlys since you already changed some of this stuff in your branch!

No worries!

I'm not sure why that test is failing... looks like it's not successfully getting the args specs?

davemfish:
> > Sorry for the future merge conflicts @emlys since you already changed some of this stuff in your branch!
> 
> No worries!
> 
> I'm not sure why that test is failing... looks like it's not successfully getting the args specs?

Yeah that was weird. It looks like it failed to download the invest binaries due to a weird ssl error. But that reminded me to update the invest version as well, so let's wait for everything to re-run.

davemfish:
> @davemfish, I like the dropdown for the n workers value! To further streamline it, instead of having the options '-1' and '0' in the dropdown, then defining them in the collapsible box, what if the definitions were right in the dropdown? So the options would be like "Synchronous (recommended)", "Multithreaded single process", "2 processes", "3 processes", ...

Yes, good suggestion. I switched the dropdown to use the same text that we have in the Qt UI. I modified the extra text, I still think it's useful to have a little extra info available than fits in the label. And I tweaked the layout to accommodate the wider dropdown content. Back to you @emlys 

settings_ui

emlys (approved):
Looks good!

PR #209, opened 2021-12-13
Task/207 fishes deprecation

davemfish:
This PR removes references to Finfish, Fisheries & Fisheries HST from the UI config file and the sampledata registry. Also updates invest to the latest `release/3.10` which includes those model deprecations.


emlys (approved):
Hooray!

PR #211, opened 2021-12-13
Alert width should not expand to fit it's content

### LINKED ISSUE #168: Sidebar should not expand to fit its content
The sidebar mostly contains static content, but the exception is the Alert that displays the final invest error message. Right now the Alert width expands to fit the error message, and in turn the sidebar width expands to fit the Alert. The main Tab Pane is pushed offscreen.

The sidebar width should have max width.

davemfish:
Fixes #168 for the edge case of very-long-text-strings-with-no-whitespace-on-which-to-break-lines.

Before image in the issue. After image:
Capture


dcdenu4 (approved):
Hey @davemfish , easy peasy!

PR #214, opened 2021-12-17
Bugfix/212 invest log regex

### LINKED ISSUE #212: bug in regex for invest logfile markup
We markup invest logfiles based on which module is doing the logging, and our pattern-matching includes a regex like`new RegExp('natcap.invest.carbon')`. We need to be careful to only find matches for strings with those literal dots.

Something like `new RegExp('natcap.invest.carbon'.replace(/\./g, '\\\\.'))` should do it.

davemfish:
This PR fixes #212 .

It also generally simplifies and cleans up this feature by moving it to a separate module. Also, when this feature was first developed, we were parsing invest logs line-by-line whereas now we parse them message-by-message, and that makes it much easier (simpler regex) to assign the correct markup to multi-line messages such as Tracebacks.

Now the markup is as follows:
* `DEBUG|INFO` messages from the invest model are bold
* `WARNING` messages from the invest model are red (not bold)
* `ERROR|CRITICAL` messages from any module are red & bold

**BEFORE:**
logging_before1
the shapely warnings are mistakenly bold because they happen to include the string `natcap/invest/coastal_vulnerability`

logging_before2
And this one is incorrectly bold red because it includes the string `AttributeError`.

**AFTER:**
logging_after
shapely messages are plain. `WARNINGS` from the invest model are elevated to red.

### DIFF
--- a/src/main/investLogMarkup.js
+++ b/src/main/investLogMarkup.js
@@ -28,5 +28,5 @@ export default function markupMessage(message, pyModuleName) {
       return sanitizeHtml(markup, allowedHtml);
     }
   }
-  return sanitizeHtml(message, allowedHtml);
+  return sanitizeHtml(message);
 }
--- a/src/main/investLogMarkup.js
+++ b/src/main/investLogMarkup.js
@@ -1,0 +1,32 @@
+import sanitizeHtml from 'sanitize-html';
+
+/**
+ * Encapsulate text in html, assigning class based on text content.
+ *
+ * @param  {string} message - from a python logger
+ * @param  {string} pyModuleName - e.g. 'natcap.invest.carbon'
+ * @returns {string} - sanitized html
+ */
+export default function markupMessage(message, pyModuleName) {
+  const escapedPyModuleName = pyModuleName.replace(/\./g, '\\.');
+  const patterns = {
+    'invest-log-error': /(ERROR|CRITICAL)/,
+    'invest-log-primary-warning': new RegExp(`${escapedPyModuleName}.*WARNING`),
+    'invest-log-primary': new RegExp(escapedPyModuleName)
+  };
+
+  const logTextTag = 'span';
+  const allowedHtml = {
+    allowedTags: [logTextTag],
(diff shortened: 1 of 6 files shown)


src/renderer/components/InvestTab/index.jsx
```diff
@@ -304,7 +304,6 @@ export default class InvestTab extends React.Component {
                   logfile={logfile}
                   executeClicked={executeClicked}
                   jobID={jobID}
-                  pyModuleName={modelSpec.pyname}
```
davemfish, line 307: `LogTab` component was not using this prop.

src/main/investLogMarkup.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  // eslint-disable-next-line
+  for (const [cls, pattern] of Object.entries(patterns)) {
+    if (pattern.test(message)) {
+      const markup = `<${logTextTag} class="${cls}">${message}</${logTextTag}>`;
+      return sanitizeHtml(markup, allowedHtml);
+    }
+  }
+  return sanitizeHtml(message, allowedHtml);
```
dcdenu4, line 31: I don't think it really matters, but if no markup has been done, do we need to pass `allowedHtml` in this instance?

dcdenu4 (approved):
Hey @davemfish , this looks good to me. I like breaking that bit of logic out into its own module. I had one question / comment but it's pretty trivial so I'm approving and will let you go ahead and merge if no changes need to be made.

src/main/investLogMarkup.js
davemfish: Good one! Change pushed.

davemfish:
> Hey @davemfish , this looks good to me. I like breaking that bit of logic out into its own module. I had one question / comment but it's pretty trivial so I'm approving and will let you go ahead and merge if no changes need to be made.

Thanks @dcdenu4 I made the change you suggested. Might as well let the tests run again.

dcdenu4 (approved):
Thanks @davemfish !

PR #218, opened 2022-01-24
Task/216 jest config

### LINKED ISSUE #216: standardize jest mocks, setup, teardown, etc
Each test file -- and each block within a file -- might do jest setup & teardown slightly differently. We should clean this up and standardize things. Specifically,

* avoid using `jest.resetAllMocks` (or `clearAll`, `restoreAll`, etc) as it will reset mocks in the `__mocks__` folder which are always meant to be in place for the entire test suite. It's not clear why a `reset` changes an implementation defined in `__mocks__`, as I think _that_ should be the original mock that is being reset _to_, but that's not how it works.
* If it ever makes sense to do those `AllMocks` operations, it can be setup in a global config so that it happens consistently for every test file
*

davemfish:
This PR fixes #216 , it standardizes our approach to jest mocking, setup and teardown.

It adds some global jest config to `package.json` that handles consistent cleanup of mocks before each test. So mock-cleanup is removed from `afterEach` and `afterAll` calls within test files. Please see the changes in `readme.md` for details on how I ended up with that new config.

Also:
* migrates from `fireEvent` to `userEvent`, since that is recommended as a better way to mimic real user interactions.
* updates to jest 27
* Better segregates tests that need a jsdom environment (`renderer` tests) from those that only need node (`main` tests).
* Some other minor changes were inspired by https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

### DIFF
--- a/__mocks__/electron.js
+++ b/__mocks__/electron.js
@@ -40,6 +40,8 @@ class MockApp extends events.EventEmitter {
       () => os.tmpdir()
     );
     this.setPath = jest.fn();
+    this.getName = jest.fn();
+    this.getVersion = jest.fn();
   }
 }
 export const app = new MockApp();
--- a/tests/jest.overrides.js
+++ b/tests/jest.overrides.js
@@ -55,6 +55,7 @@ if (global.window) {
   // which is used in the renderer by app.jsx and InvestJob.js.
   // A better solution is to avoid reliance on that node module
   // in the renderer process.
+  // https://github.com/natcap/invest-workbench/issues/60
   global.window.crypto = {
     getRandomValues: () => {
       return [crypto.randomBytes(4).toString('hex')];
--- a/tests/jest.overrides.js
+++ b/tests/jest.overrides.js
@@ -2,11 +2,11 @@ const crypto = require('crypto');
 
 // Cause tests to fail on console.error messages
 // Taken from https://stackoverflow.com/questions/28615293/is-there-a-jest-config-that-will-fail-tests-on-console-warn/50584643#50584643
-let error = console.error;
-console.error = function (message) {
-  error.apply(console, arguments); // keep default behaviour
-  throw (message instanceof Error ? message : new Error(message));
-};
+// let error = console.error;
+// console.error = function (message) {
+//   error.apply(console, arguments); // keep default behaviour
+//   throw (message instanceof Error ? message : new Error(message));
+// };
 
 // I've found this obfuscates error messages sometimes:
 /* With this override:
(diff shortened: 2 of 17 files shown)


tests/jest.overrides.js
```diff
@@ -2,11 +2,11 @@ const crypto = require('crypto');
...
-let error = console.error;
-console.error = function (message) {
-  error.apply(console, arguments); // keep default behaviour
-  throw (message instanceof Error ? message : new Error(message));
-};
+// let error = console.error;
+// console.error = function (message) {
+//   error.apply(console, arguments); // keep default behaviour
+//   throw (message instanceof Error ? message : new Error(message));
+// };
```
emlys, line 9: Just to double-check, you mean to keep this commented out here, right?

tests/main/main.test.js
```diff
@@ -115,7 +106,7 @@ describe('extractZipInplace', () => {
   let file2Path;
   let doneZipping = false;
 
-  beforeEach(() => {
+  beforeEach((done) => {
```
emlys, line 109: Could you explain what's going on with `done` here? I couldn't find anything about it in the jest docs

tests/renderer/setuptab.test.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import {
   createEvent, fireEvent, render, waitFor, within
```
emlys, line 4: Are there remaining uses of `fireEvent` that could be changed to use `userEvent`?

__mocks__/electron.js
```diff
@@ -40,6 +40,8 @@ class MockApp extends events.EventEmitter {
       () => os.tmpdir()
     );
     this.setPath = jest.fn();
+    this.getName = jest.fn();
+    this.getVersion = jest.fn();
```
emlys, line 44: Are these needed because of the switch to `jsdom` environment?

emlys (commented):
@davemfish this is great. That readme is super helpful. Just had a few questions, most likely no changes needed.

tests/jest.overrides.js
davemfish: I go back and forth on it. But yeah I'm happy enough to leave it commented. It's sometimes useful to uncomment it.

tests/main/main.test.js
davemfish: Here's the docs: https://jestjs.io/docs/asynchronous

And then from https://jestjs.io/docs/setup-teardown:
> beforeEach and afterEach can handle asynchronous code in the same ways that tests can handle asynchronous code - they can either take a done parameter or return a promise.

In this case there is async stuff happening in the setup; we need to wait for the `on('close'...` handler to finish. But there is no Promise we're waiting for, so `done` seemed like a way to control things.

tests/renderer/setuptab.test.js
davemfish: Yeah, the remaining uses are for the drag-and-drop tests. There was no clear way to trigger those events with `userEvent`. Here's a convo about it: https://github.com/testing-library/user-event/issues/440

__mocks__/electron.js
davemfish: Hmm, I can't quite recall why these are needed now and they weren't before. I do remember adding these because a test was triggering these API calls and they're not available outside an electron runtime. So these mocks are necessary, I just can't explain why we didn't need them here earlier.

davemfish:
> @davemfish this is great. That readme is super helpful. Just had a few questions, most likely no changes needed.

Thanks @emlys ! As you might be able to tell from readme, I understand this better than I did before, but there are still some minor mysteries. I think I addressed your comments. No changes pushed.

tests/main/main.test.js
emlys: thanks!

emlys (approved):
Thanks for the explanations! It makes sense enough to me and we don't need to solve all the mysteries 😄

PR #219, opened 2022-01-27
optimize calls to invest validation

### LINKED ISSUE #144: optimize calls to invest validation
The change handler for model input forms triggers calls to invest validation - via the flask app's `/validate` endpoint. In my experience that function is fast enough to run on every keystroke without causing problems. But there are a couple reasons why it makes sense to optimize by limiting the number of `/validate` requests we make.

* We recently noticed a problem related to the amount of python logging emitted during those calls. So reducing the number of calls might alleviate that problem. https://github.com/natcap/invest/issues/563
* My experience that everything runs fast enough now might not hold for all users with fewer compute resources
* There's a fair amount of variation in the runtime of each model's `validate` function and I'm sure I haven't tested all models to claim that all are fast enough as-is.
* There's not much value in validating on *every* keystroke. We're not doing anything like auto-complete that would benefit from updating with every new character. We really only need to validate the intended user-input, not the partially complete intended user-input.

What we're talking about here is called "debouncing". And there's a number of ways to do it. https://stackoverflow.com/questions/23123138/perform-debounce-in-react-js

Importantly, we don't want to debounce the change handler itself - we want the text to update in response to every keystroke - we want to debounce the call to `investValidate`.

### LINKED ISSUE #182: invest validation is requested when fields are touched but not changed
Just toggling the cursor between input fields triggers calls to invest's validation endpoint, even when none of the input values have actually changed. Those calls are unnecessary.

davemfish:
This fixes #144 and fixes #182 

It adds a `debouncedValidate` method to delay requests to invest validation until there is a pause in data entry. I set this at 200ms, which seems like a decent compromise between getting very responsive validation, but not validating too often as someone types or lays on the backspace key.

There's also a new method for handling focus on an input differently from handling changes to the input value.

### DIFF
--- a/src/renderer/components/SetupTab/ArgInput/index.jsx
+++ b/src/renderer/components/SetupTab/ArgInput/index.jsx
@@ -115,6 +115,7 @@ export default class ArgInput extends React.PureComponent {
       enabled,
       handleBoolChange,
       handleChange,
+      handleFocus,
       inputDropHandler,
       isValid,
       selectFile,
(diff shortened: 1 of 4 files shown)


dcdenu4 (approved):
Hey @davemfish this looks like a nice implementation. I don't have any suggestions or questions.

